PR #3, opened 2015-03-07
- one more step for the guide

damianszczepanik:
- added new small but very old repos to the list


romani:
@damianszczepanik , it was inaccurate from me to apply that ....

> cd sevntu-checkstyle

Where is not folder "sevntu-checkstyle" , so we can not  provide commands without switching user to another project.

> mvn clean install

Why that is required ? as deploy-all.sh should do all of that.

damianszczepanik:
It complained that check is not known so I `mvn clean install` and `deploy-all.sh` again. Then the problem was fixed and report was generated.

PR #6, opened 2015-03-22
- Support to pull given commit so each execution of launch.sh validates the same sources

damianszczepanik:
- added ant repository that is quite old and good candidate when looking for violations

### DIFF
--- a/checkstyle-tester/README.md
+++ b/checkstyle-tester/README.md
@@ -8,6 +8,11 @@ Launch command for testing against your custom config:
 ./launch.sh -Dcheckstyle.config.location=my_check.xml
 ```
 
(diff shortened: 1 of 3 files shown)


romani:
1) 
> ./deploy-all.sh

why not ```./deploy.sh --maven``` ? See second "ATTENTION: " in that file.

2) can we make commit-ref as optional ? in some cases user do not care what version of code to test.

damianszczepanik:
1. It did not work for me but I will try again, probably my mistake
2. The reason of this is that I thought about requirements in https://github.com/checkstyle/checkstyle/issues/555 but sounds like having this optional is good idea

will update PR

romani:
thanks a lot, please put comment to PR  after code change to notify me that you completed the change, in other case I miss event and PR is hanging.

PR #17, opened 2015-08-22
FileLengthCheck was commented out in qulice-checks

baratali:


romani:
why we can not use our suppression in qulice-checks.xml, it even could be used by URL reference.

PR #43, opened 2015-12-06
Pull #43: [release-notes-xdoc-builder] implement Builder patter for CliOptions class

MEZk:
@romani 
Please review.


romani:
I do not thinks that prefix "set" for Builder methors is good idea, it is usually avoided to simplify code.

MEZk:
@romani 
Corrected

PR #45, opened 2015-12-07
Pull #45:  [releasenotes-xdoc-builder] workaround to fix first line indentation

MEZk:


romani:
Can we fix template ? If not please provide a reasons.

MEZk:
@romani 
We cannot fix template itself. Thymeleaf generates out file where indentation of the first line is incorrect. It is not a problem of the template. It is a problem of thymeleaf engine.

romani:
Can you provide link to issue against engine ? Before we give up to use thymeleaf and switch to freemarker.

our java code should not any updates to output.

MEZk:
@romani 
I'll try to ivestigate the problem deeper. Maybe we can switch wersion from 2+ to 3+ after release of thymeleaf. 

http://mvnrepository.com/artifact/org.thymeleaf/thymeleaf-spring3/3.0.0.BETA01

PR #47, opened 2015-12-17
[release-notes-xdoc-builder] minor fixes and refactoring

MEZk:
1) Themyleaf version was updated to 3.0.0.BETA01 and as a result problem with indentation of the first line in output file has been resolved.
2) Made refactoring of TemplateProcessor to use more abstract class as template resolver and exclude unnecessary instantiation of String object.
3) Reformated thymeleaf template to resolve the problem with blank lines in output file which are violated by Checkstyle. I tried to save readability of template.
4) Please rename https://github.com/checkstyle/checkstyle/commit/89cee419943977af6e70264a10c669a4c29bf830 if you want it to be excluded from release notes.


romani:
There is no way to rename commit, commits with "conf:" prefix should be skipped by builder

MEZk:
@romani 
Hm, what about ```git rebase -i HEAD~N```,
where N is a number of commits after the target commit?

PR #48, opened 2015-12-22
checkstyle-tester launcher has been updated to detect OS and use appropriate sed flag

sabaka:
Should fix problem with sed on MacOS


romani:
problem with report parsing still present an confirmed to exists.

PR #53, opened 2016-01-24
Changes for circleCI

Vladlis:
empty_file extension is changed on .java to let checkstyle-tester run on no projects without exceptions caused by absence of generated folders


romani:
I see only delation of file, and no new java file

Vladlis:
I guess Github won't show it as a new file, but strange that the extension didn't change. Anyways, it is not needed anymore.

PR #87, opened 2016-04-25
checkstyle-web-tester: added first release

rnveach:
Not to be confused with `checkstyle-tester` for regression, web tester is just a simple web interface for checkstyle. Users can supply config, code, and CS version and run it. Links are generated for the file and config for easy sharing. You can select any installed CS release for regression.
Textarea fields have line numbers so its easy to see what line violations point to.

**Live version may be found here: http://rveach.no-ip.org/checkstyle/checkstyle.php**

You may use this as you wish, but I can't guarantee uptime or how long files will be saved, if you use that feature. Max 5kb limit for all files.
Releases and nightlies have to be manually installed.

### DIFF
--- a/checkstyle-web-tester/README.md
+++ b/checkstyle-web-tester/README.md
@@ -1,0 +1,34 @@
+About
+=====
+
+CheckStyle Web Tester is a web utility for regression and testing simple scenarios using checkstyle utility.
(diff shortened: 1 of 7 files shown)


romani:
please add this link to README.md nobody will find this PR.

PR #94, opened 2016-06-18
checkstyle-tester: Add Apache Struts into projects-to-test-on.properties

MEZk:
Apache Struts 2 web framework is a free open-source solution for creating Java web applications. Apache team no longer supports Struts 1, but Apache Struts 2 contains a lot of legacy code. The code is also packed with annotations (Injections, ets) and will be good choice for testing Checkstyle's annotation checks.

https://github.com/apache/struts


romani:
as it so good, lets add it to 2 other property files and activate it Travis or CircleCI to run always.
in other case it could be forgotten.

PR #106, opened 2016-07-23
minor: synchronized and update projects properties

rnveach:
Requested because of PR https://github.com/checkstyle/checkstyle/pull/3357#issuecomment-234233747

I synchronized all 3 property files, so they all contain the same information and easier to do merges/compares with.
I replaced all github protocols with git.
I updated Orekit with the latest commit.

I didn't see any other discrepancies between the properties and wrecker/travis configurations.


romani:
thanks a lot !

Issue #73, opened 2016-03-18
Automate social posts (SF, RSS, twitter, g+, maillist)


romani:
Assigned to @Vladlis .

Right now I do this manually, this should be automated.

Eventually it be launched on Travis automatically. Preferred languages, shell + java(if required) .

Vladlis:
@romani 
1) the format for g+ and twitter is not obvious. I propose the following: https://plus.google.com/108692338295483568100/posts/K3pCdYH5wH1

2) I do not see any publications about releases in [maillist](https://groups.google.com/forum/#!forum/checkstyle) - will we start it? If so, then what format will we use - the same as in RSS and SF?

romani:
> 1)

yes, that format is expected, just link and statistics on amount of updates.

> I do not see any publications about releases in mail-list - will we start it?

Special mail-list for announcements - https://groups.google.com/forum/#!forum/checkstyle-announce
Format the same as for Sourceforge.

romani:
we need options:
- generateAll (all posts)
- generateXXXX (specific post)
- publishAll (publish all posts)
- publishXXXX (publish specific post)
- loginXXXX/pwdXXXX for each post to publish

such options are required to skip some publishing if it is become unstable or problematic.

Vladlis:
@romani 
Please, review the last commit from this issue to confirm that I'm on the right way.

romani:
sorry for long delay.

1)
> CliOptions

it is general it s good. There is only one bad side of this implementation of Builder is that `build()` method will return the same instance of CliOptoin of class, rather than new instance.
```
Builder builder = CliOptions.newBuilder() 
CliOption o = builder.releaseNotes(commandLine.getOptionValue(OPTION_R))
           .outputLocation(commandLine.getOptionValue(OPTION_O))
            ...
           .build();

CliOption o1 = builder.releaseNotes(commandLine.getOptionValue(OPTION_R))
           .outputLocation(commandLine.getOptionValue(OPTION_O))
            ...
           .build();

```

builder will return the same instance, but this is not obvious. Builder should always return new instance. User of this builder should trust a name and do not recheck how this builder is implemented.
I know that we have form of builder in checkstyle - it was my bad to allow it to appear.

romani:
@Vladlis ,

so as we discussed before , Social posts is just another release notes generation,
the only nuances:
1) xdoc + social generation should be done before "mvn release:prepare"
[...]

PR #110, opened 2016-08-07
Issue #73: Add support for xdoc publication (commit) and fix IntelliJ IDEA inspections violations

Vladlis:


romani:
merged as FF.

@Vladlis , please  ad such publish process to checkstyle's CI, to make sure it works on each commit and will never be surprise-failure during release. you can make a `head -n 100 releasenotes.xml` command to print result file to build output and allow human verify result.

PR #124, opened 2016-11-10
Issue #73: Update an option and add examples of publication commands

Vladlis:
Issue #73

### DIFF
--- a/releasenotes-builder/README.md
+++ b/releasenotes-builder/README.md
@@ -18,9 +18,10 @@ Jar file which includes all required dependencies will be located at
 ## Command line usage
 ```
 java -jar releasenotes-builder-1.0-all.jar -localRepoPath <arg> -startRef <arg> [-endRef <arg>] \
-     -releaseNumber <arg> [-outputLocation <args>] [-authToken <arg>] \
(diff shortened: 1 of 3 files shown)


romani:
@Vladlis , please rename authToken to githubAuthToken to make it clear what kind of token is required for builder.

Vladlis:
@romani,
looks like the error from TC should be warning
https://github.com/checkstyle/contribution/blob/master/releasenotes-builder/config/intellij-idea-inspections.xml#L1743

romani:
1) suppression of DesignForExtension is not a solution, @MEZk updated this Check not for ignoring. Please consult MEZk on how to fix.

2)
> looks like the errors from TC should be warnings

it is already a warning, but TC do not care about severity, all violations fail the build. To disable rule we use `... enabled="false" ...`.

Need to be resolved, as we use the same rule in main project too.

romani:
CIs was red even before your PR.
Fixed.
@Vladlis , please rebase your changes on latest code. ALL CIs should be green , if not , it caused by your PR - please fix.

Vladlis:
Rebased

Issue #51, opened 2015-12-31
automate checkstyle-tester for diff report generation


romani:
checkstyle-tester should have a mode to completely automatically generate diff report base on:
- latest binaries of chekcstyle (we could use our SNAPSHOT maven repo)
- "link" to patch (or some other way defining new binaries)
- config to run 

result should be ready to upload to github.io folder with report.

MEZk:
@romani 
We can improve the tool (launch_diff.sh) designed by @rnveach .
I am able to generate diff reports with the tool. However, it needs to receive the following parameters which now are specified directly in the script:

MINIMIZE
CONTACTSERVER
CHECKSTYLE_DIR
TESTER_DIR
EXTRA_DIR
FINAL_RESULTS_DIR
DIFF_JAR
PULL_REMOTE

>latest binaries of chekcstyle (we could use our SNAPSHOT maven repo)

Why can't we run mvn install on the latest master? If we use SNAPSHOT maven repo it will require additional logic to grap the latest SNAPSHOT.

>"link" to patch (or some other way defining new binaries)

There is no need in passing the link. With @rnveach 's script it is generated and used automatically. We can generate before/after reports in the root of checkstyle-tester (for example in checkstyle-tester/reports-tmp), generate diff report in checkstyle-tester/diff-report and remove reports-tmp. So, final reports will be located at checkstyle-tester/diff-report.

>config to run

Now the script has an option which allows to specify Checkstyle config for the tester.

romani:
> We can improve the tool (launch_diff.sh) designed by @rnveach .

yes reusage of exisinng is good.
The only concern for me is that usage of shell there might be not a good idea. Big shell files are hard to maintain.
@rnveach , does it make sense to rewrite it to groovy (java scripting language) ?
Most CIs support it, it should not be a problems for contributors to run groovy too. Probably we could make it work on Windows too.

> However, it needs to receive the following parameters which now are specified directly in the script

all that variables should be based on filesystem where clone for contribution repo was done. All additional clones generations should happen under some inner folder.
It might be good to be all that variables to be customizable to let tune it, but it should work by default without any extra steps/configurations.

> Why can't we run mvn install on the latest master?

no problem, it was just an idea, use what ever is convenient.

> There is no need in passing the link. 

by link I meant some indentifier of path - "username" and "branch". Example - https://github.com/checkstyle/checkstyle/blob/master/fast-forward-merge.sh#L14

[...]

PR #125, opened 2016-11-21
Issue #51: Implement Checkstyle report generation script in Groovy

MEZk:
Implemented Checkstyle report generation script in Groovy.

Report folder diff (shell vs Groovy implementation):

cmp

script cli args example:  ```groovy launch -c my_check.xml -p projects-to-test-on.properties```

Should I update README to explain how to use the new script?

### DIFF
--- a/checkstyle-tester/launch.groovy
+++ b/checkstyle-tester/launch.groovy
@@ -1,0 +1,275 @@
+import java.nio.file.Files
+import java.nio.file.Paths
+import java.util.regex.Pattern
+
+import static java.lang.System.err
+
+if (areValidCliArgs(args)) {
+    def projectsToTestOn = args[0]
+    def checkstyleCfg = args[1]
+    generateCheckstyleReport(projectsToTestOn, checkstyleCfg)
+}
+else {
+    throw new IllegalArgumentException('Error: invalid command line arguments!')
+}
+
+def areValidCliArgs(args) {
+    def valid = true
+    if (args == null || args.length == 0) {
+        valid = false
+    }
+    else {
+        def validArgsCount = 2
+        if (args.length == validArgsCount) {
+            def projectsToTestOnFile = new File(args[0])
+            def checkstyleCfgFile = new File(args[1])
+            if (!projectsToTestOnFile.exists()) {
+                err.println "Error: file ${projectsToTestOnFile.getName()} does not exist!"
+                valid = false
+            }
+            else if (!checkstyleCfgFile.exists()) {
+                err.println "Error: file ${checkstyleCfgFile.getName()} does not exist!"
+                valid = false
+            }
+        }
+        else {
+            err.println 'Error: wrong number of command line arguments!'
+            valid = false
+        }
+    }
+    return valid
+}
+
+def generateCheckstyleReport(projectsToTestOn, checkstyleConfig) {
+    println 'Testing Checkstyle started'
+    def projectsToTestOnFile = new File(projectsToTestOn)
+    def projects = projectsToTestOnFile.readLines()
+
+    def REPO_NAME_PARAM_NO = 0
+    def REPO_TYPE_PARAM_NO = 1
+    def REPO_URL_PARAM_NO = 2
+    def REPO_COMMIT_ID_PARAM_NO = 3
+    def REPO_EXCLUDES_PARAM_NO = 4
+    def FULL_PARAM_LIST_SIZE = 5
+
+    def srcDir = "src/main/java"
+    def reposDir = 'repositories'
+    def reportsDir = 'reports'
+    createWorkDirsIfNotExist(srcDir, reposDir, reportsDir)
+
+    def targetDir = 'target'
+    def siteDir = "$targetDir/site"
+
+    projects.each {
+        project ->
+            if (!project.startsWith('#') && !project.isEmpty()) {
+                def params = project.split('\\|')
+                def repoName = params[REPO_NAME_PARAM_NO]
+                def repoType = params[REPO_TYPE_PARAM_NO]
+                def repoUrl = params[REPO_URL_PARAM_NO]
+                def commitId = params[REPO_COMMIT_ID_PARAM_NO]
+                def excludes = ''
+                if (params.length == FULL_PARAM_LIST_SIZE) {
+                    excludes = params[REPO_EXCLUDES_PARAM_NO]
+                }
+
+                cloneRepository(repoName, repoType, repoUrl, commitId, reposDir)
+                copyDir("$reposDir/$repoName", "$srcDir/$repoName")
+                runMavenExecution(srcDir, excludes, checkstyleConfig)
+                postProcessCheckstyleReport(siteDir)
+                deleteDir("$srcDir/$repoName")
+                moveDir(targetDir, "$reportsDir/$repoName")
+            }
+    }
+}
+
+def createWorkDirsIfNotExist(srcDirPath, repoDirPath, reportsDirPath) {
+    def srcDir = new File(srcDirPath)
+    if (!srcDir.exists()) {
+        srcDir.mkdirs()
+    }
+    def repoDir = new File(repoDirPath)
+    if (!repoDir.exists()) {
+        repoDir.mkdir()
+    }
+    def reportsDir = new File(reportsDirPath)
+    if (!reportsDir.exists()) {
+        reportsDir.mkdir()
+    }
+}
+
+def cloneRepository(repoName, repoType, repoUrl, commitId, srcDir) {
+    def srcDestinationDir = "$srcDir/$repoName"
+    if (!Files.exists(Paths.get(srcDestinationDir))) {
+        def cloneCmd = getCloneCmd(repoType, repoUrl, srcDestinationDir)
+        println "Cloning $repoType repository '$repoName' to $srcDestinationDir folder ..."
+        executeCmd(cloneCmd)
+        println "Cloning $repoType repository '$repoName' - completed\n"
+    }
+
+    if (commitId && commitId != '') {
+        def lastCommitSha = getLastCommitSha(repoType, srcDestinationDir)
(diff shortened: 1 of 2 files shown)


rnveach:
@MEZk I assume this is for issue 51, not 52?
I think we should have a readme for it. I'm not sure if it should be in the main readme or its own, as we will want to directly link to it for newcomers.
Also remember some people don't know groovy, like me, and they may be more familar with Java since that is what the main project is written in.

MEZk:
@rnveach 
Yes, it is for #51 ) Fixed.

>I think we should have a readme for it. I'm not sure if it should be in the main readme or its own, as we will want to directly link to it for newcomers.

@romani 
What is your opinion?

>Also remember some people don't know groovy, like me, and they may be more familar with Java since that is what the main project is written in.

Groovy is similar to java and run in JVM. It is more readable and easy to support for newbies. I also do not know Groovy at all, but my experience in Java made it easy for me to implement the script. Pure Java is to heavy for scripting, and even though Groovy is running in JVM, it has lightweight scripting support.

rnveach:
@MEZk 
I'm confused. Isn't this issue suppose to be a full automation script of regression like my script, `launch_diff.sh`?
This just looks more like `launch.sh`. Are we doing both in groovy?

romani:
I presume that this is a just a first milestone on #51 - change launch_sh to groovy. 
Next milestone is finally resolve #51 and have groovy script to generate diff report too.
@MEZk , please confirm your intentions.

MEZk:
@rnveach @romani 
This is just the first step. Next step will be the Groovy script which uses launch.groovy to generate reports and get diff based on them. After the merge of this PR I am going to change CIs yml matrix to run the new script in Groovy.

rnveach:
> Next step will be the Groovy script which uses launch.groovy to generate reports and get diff based on them.

@romani @MEZk 
Just to be clear, `launch_diff` doesn't call `launch`.

Each project gets its own separate maven run, which is not what happens in `launch`. In `launch` everything is packed into one directory and JXR is run in one pass.
There are very good reasons for why my script didn't follow suite.

Putting too many projects in one run, requires too much memory. We also need more memory for violations and some checks like indentation put out alot of violations. I was always required to give indentation 4 gigs of memory to complete all projects, even though each project was run separate.
Also if you pack all the projects together, you won't get 1 HTML report per project unless you modify the resulting `checkstyle-result.xml` that the diff report needs. All the projects will be combined into one HTML, and could be huge and slow to load in the browser.

Another thing my script did, was verify that the checkstyle version installed matches the version specified in the [POM](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/pom.xml#L15).
If they aren't in synch, the checkstyle version run is usually an old version and there is no clear way to notice this unless you are expecting differences. This caused almost quite a few invalid regression reports for me until I found and added the check.
You can't really do this unless it is one full suite or passing data back and forth.

I am not saying we can't convert `launch` and use it, but I don't think it can be used for the full regression suite.

MEZk:
@rnveach @romani 

>Each project gets its own separate maven run, which is not what happens in launch

We can add an option which specifies the project to generate report for. It will have higher priority than "-p"  and run the script only for one project.

>Putting too many projects in one run, requires too much memory. 

Yes. CIs will not allow us to use so much memory. And it is a problem, but if we split report generation into steps (each for one project) it will help.

romani:
This is first step to reusable code. Eventually groovy script will do all that possible by both of shell scripts. And we will have one script that we need to support.

@rnveach , please join me in review of this PR.

checkstyle-tester/launch.groovy
```diff
@@ -0,0 +1,191 @@
...
+                def repoType = params[REPO_TYPE_PARAM_NO]
+                def repoUrl = params[REPO_URL_PARAM_NO]
+                def commitId = params[REPO_COMMIT_ID_PARAM_NO]
+                if (params.length == FULL_PARAM_LIST_SIZE) {
+                    excludes += ',' + params[REPO_EXCLUDES_PARAM_NO]
+                }
+                cloneRepository(repoName, repoType, repoUrl, commitId, srcDir)
+            }
+    }
+    runMavenExecution(srcDir, excludes, checkstyleConfig)
```
romani: please be consistent in sequence of arguments. I like this sequence , so please fix "+def generateCheckstyleReport(checkstyleConfig, srcDir, projectsToTestOn) {"
to be 
"def generateCheckstyleReport(srcDir, projectsToTestOn, checkstyleConfig) {"

checkstyle-tester/launch.groovy
```diff
@@ -0,0 +1,191 @@
...
+    println "Running Checkstyle on $srcDir ... with excludes $excludes"
+    def mvnSite = "mvn -e --batch-mode site -Dcheckstyle.config.location=$checkstyleConfig -Dcheckstyle.excludes=$excludes -DMAVEN_OPTS=-Xmx3024m"
+    executeCmd(mvnSite)
+    println "Running Checkstyle on $srcDir - finished"
+}
+
+def postProcessCheckstyleReport() {
+    println 'linking report to index.html'
+    new File('target/site/index.html').renameTo 'target/site/_index.html'
+    Files.createLink(Paths.get("target/site/index.html"), Paths.get("target/site/checkstyle.html"))
```
romani: I am not sure this will work for Windows, @rnveach please confirm.
we can use "mv"/"rename".

checkstyle-tester/launch.groovy
```diff
@@ -0,0 +1,191 @@
...
+
+    removeNonReferencedXrefFiles()
+    removeEmptyDirectories()
+}
+
+def removeNonReferencedXrefFiles() {
+    println 'Removing non refernced xref files in report ...'
+
+    def linesFromIndexHtml = Files.readAllLines(Paths.get('target/site/index.html'))
```
romani: make a function from this block.

checkstyle-tester/launch.groovy
```diff
@@ -0,0 +1,191 @@
...
+                    && 'overview-summary.html' != fileName) {
+                fileObj.delete()
+            }
+    }
+}
+
+def removeEmptyDirectories() {
+    // we use unix commands to avoid problems with concurrency modification of collection
+    // (removing and iteration over collection) in Groovy
+    executeCmd('find . -type d -empty -delete', new File('./target/site/xref'))
```
romani: non Windows friendly. But might be ok for first iteration.

checkstyle-tester/launch.groovy
```diff
@@ -0,0 +1,191 @@
...
+}
+
+def executeCmd(cmd, dir = new File("./")) {
+    def proc = cmd.execute(null, dir)
+    proc.consumeProcessOutput(System.out, System.err)
+    proc.waitFor()
+    if (proc.exitValue() != 0) {
+        println "Error: ${proc.err.text}!"
+        System.exit(-1)
```
romani: please throw exception there.

checkstyle-tester/launch.groovy
rnveach: It should work. http://stackoverflow.com/questions/12509132/how-to-force-java-7-to-create-directory-symlinks-on-windows-instead-of-file-s?rq=1
I've played with another utility that creates symlinks on Windows 7, so it is supported, just not easily accessible.
http://www.howtogeek.com/howto/16226/complete-guide-to-symbolic-links-symlinks-on-windows-or-linux/

rename in windows is: `ren`. mv in windows is: `move`.

rnveach:
Github is weird. I finished my review now, but because some were made long ago, it put the 2 review items and comments up near the top of this thread.

MEZk:
@romani

> I think it make sense to use "public static void main(args) {"

Done.

> is it really required ?

No. Removed.

> please be consistent in sequence of arguments. I like this sequence , so please fix "+def generateCheckstyleReport(checkstyleConfig, srcDir, projectsToTestOn) {"
to be
"def generateCheckstyleReport(srcDir, projectsToTestOn, checkstyleConfig) {"

Put srcDir and other paths which are used by generateCheckstyleReport inside generateCheckstyleReport

> make a function from this block.

Done.

> non Windows friendly.

Done. Implemented recursive method to delete empty directories.

MEZk:
@rnveach

> Since we are downloading the repos and not cleaning them out, for long term users who will use this for years to come, you should put in some git fetch and reset commands like launch_update. Just so we are always running on the project's latest code, unless a specific commit is set.

Contributor can use 'master'/'HEAD' for git or 'default' for mercurial tags in projects-to-test-on.properties to force the script to download the lates sources. Maybe it will be better to add description in README.

> I recommend putting all downloaded repos in a separate directory not in src/main/java and then copying them over for the run.
Any new directories added will have to be added to gitignore.

Done.

> Will I have to specify the one project to run? I would dislike having to redo the loading/parsing of the projects to test on outside of this script. Can't we just make this script run all the projects in a single/separate run?

Yes. Done.

> do we need to sanitize directory name $repoName for invalid characters?
Windows users have some characters that can't be in directory names.
Can wait when we make this more windows friendly.

Paths.get() will throw IllegalPathException in case the path is not correct. 

> Maybe verify if current commit id is equal to commitId? Then we can avoid doing a useless reset.

Done.

MEZk:
@romani @rnveach 
Now the script copies the full content of 'target' directory ('site' folder, checkstyle-result.xml, checkstyle-checker.xml) into reports directory. However, diff tool requires only checkstyle-result.xml. Nevertheless, I decided to leave the whole content as contributer may need to analyze report just 'after' fix without diff. But the size of such reports may be too big. What is your opinion? Should we have an option to generate just checkstyle-result.xml?

romani:
Diff report and simple html report are completely different modes. Both are required for now, in future smth could change as checkstyle could generate html report itself.

romani:
It is fine for first step , very good step forward.

MEZk:
@romani 
Did you merge changes? I do not see the script in master branch.

romani:
Merged.
Sorry.

PR #136, opened 2016-12-11
Issue #51: Add diff report generation automation script in Groovy

MEZk:
#51

@romani @rnveach 

1) Reports before fix will be generated into /reports/before_fix/{project_name} directory
2) Reports after fix will be generated into /reports/after_fix/{project_name} directory
3) Diff report will be generated into /reports/diff/{project_name} directory
4) Summary index.html will be in /reports/diff directory

Please, review

P.S.
I'll add documentation in README.md later when we agree on implementation and cli args.

### DIFF
--- a/checkstyle-tester/diff.groovy
+++ b/checkstyle-tester/diff.groovy
@@ -1,0 +1,264 @@
+import java.nio.file.Paths
+
+import static java.lang.System.err
+
+static void main(String[] args) {
+    def cli = new CliBuilder(usage:'groovy diff.groovy [options]')
+    cli.with {
+        r(longOpt: 'localGitRepo', args: 1, argName: 'localGitRepo', 'Path to local git repository')
+        b(longOpt: 'baseBranch', args: 1, argName: 'baseBranch', 'Base branch name. Default is master')
+        p(longOpt: 'patchBranch', args: 1, argName: 'patchBranch', 'Name of the patch branch in local git repository')
+        c(longOpt: 'checkstyleCfg', args: 1, argName: 'checkstyleCfg', 'Path to checkstyle config file')
+        l(longOpt: 'projectsToTestOn', args: 1, argName: 'projectsToTestOn', 'Path to file which contains projects to test on')
+    }
+    def options = cli.parse(args)
+
+    if (areValidCliArgs(options)) {
+        def localGitRepo = new File(options.localGitRepo)
+        if (hasUnstagedChanges(localGitRepo)) {
+            throw new IllegalStateException("Error: git repository ${localGitRepo.getPath()} has unstaged changes!")
+        }
+
+        def baseBranch = options.baseBranch
+        if (!baseBranch) {
+            baseBranch = 'master'
+        }
+
+        def patchBranch = options.patchBranch
+        def projectsToTestOn = options.projectsToTestOn
+        def checkstyleCfg = options.checkstyleCfg
+
+        def reportsDir = 'reports'
+        def masterReportsDir = "$reportsDir/$baseBranch"
+        def patchReportsDir = "$reportsDir/$patchBranch"
+
+        def tmpReportsDir = 'tmp_reports'
+        def tmpMasterReportsDir = "$tmpReportsDir/$baseBranch"
+        def tmpPatchReportsDir = "$tmpReportsDir/$patchBranch"
+
+        def diffDir = "$reportsDir/diff"
+
+        // Delete work directories to avoid conflicts with previous reports generation
+        if (new File(reportsDir).exists()) {
+            deleteDir(reportsDir)
+        }
+        if (new File(tmpReportsDir).exists()) {
+            deleteDir(tmpReportsDir)
+        }
+
+        generateCheckstyleReport(localGitRepo, baseBranch, checkstyleCfg, projectsToTestOn, tmpMasterReportsDir)
+        generateCheckstyleReport(localGitRepo, patchBranch, checkstyleCfg, projectsToTestOn, tmpPatchReportsDir)
+        deleteDir(reportsDir)
+        moveDir(tmpReportsDir, reportsDir)
+        generateDiffReport(reportsDir, masterReportsDir, patchReportsDir, checkstyleCfg)
+        generateSummaryIndexHtml(diffDir)
+    }
+    else {
+        throw new IllegalArgumentException('Error: invalid command line arguments!')
+    }
+}
+
+def areValidCliArgs(options) {
+    def valid = true
+    if (options == null) {
+        valid = false
+    }
+    else {
+        if (options.checkstyleCfg
+                && options.localGitRepo
+                && options.patchBranch
+                && options.projectsToTestOn) {
+            def localGitRepo = new File(options.localGitRepo)
+            def patchBranch = options.patchBranch
+            def baseBranch = options.baseBranch
+            if (!isValidGitRepo(localGitRepo)
+                    || !isExistingGitBranch(localGitRepo, patchBranch)
+                    || (baseBranch && !isExistingGitBranch(localGitRepo, baseBranch))) {
+                valid = false
+            }
+        }
+        else {
+            err.println 'Error: wrong number of command line arguments!'
+            valid = false
+        }
+    }
+    return valid
+}
+
+def isValidGitRepo(gitRepoDir) {
+    def valid = true
+    if (gitRepoDir.exists() && gitRepoDir.isDirectory()) {
+        def gitStatusCmd = "git status".execute(null, gitRepoDir)
+        gitStatusCmd.waitFor()
+        if (gitStatusCmd.exitValue() != 0) {
+            err.println "Error: \'${gitRepoDir.getPath()}\' is not a git repository!"
+            valid = false
+        }
+    }
+    else {
+        err.println "Error: \'${gitRepoDir.getPath()}\' does not exist or it is not a directory!"
+        valid = false
+    }
+    return valid
+}
+
+def isExistingGitBranch(gitRepo, branchName) {
+    def exist = true
+    def gitRevParseCmd = "git rev-parse --verify $branchName".execute(null, gitRepo)
+    gitRevParseCmd.waitFor()
+    if (gitRevParseCmd.exitValue() != 0) {
+        err.println "Error: git repository ${gitRepo.getPath()} does not have a branch with name \'$branchName\'!"
+        exist = false
+    }
+    return exist
+}
+
+def hasUnstagedChanges(gitRepo) {
+    def hasUnstagedChanges = true
+    def gitStatusCmd = "git status".execute(null, gitRepo)
+    gitStatusCmd.waitFor()
+    def gitStatusOutput = gitStatusCmd.text
+    if (gitStatusOutput.contains("nothing to commit, working directory clean")) {
+        hasUnstagedChanges = false
+    }
+    return hasUnstagedChanges
+}
+
+def getCheckstyleVersionFromPomXml(pathToPomXml, xmlTagName) {
+    def pomXmlFile = new File(pathToPomXml)
+    def checkstyleVersion
+    pomXmlFile.eachLine {
+        line ->
+            if (line.matches("^.*<$xmlTagName>.*-SNAPSHOT</$xmlTagName>.*")) {
+                checkstyleVersion = line.substring(line.indexOf('>') + 1, line.lastIndexOf('<'))
+                return true
+            }
+    }
+    if (checkstyleVersion == null) {
+        throw GroovyRuntimeException("Error: cannot get Checkstyle version from $pathToPomXml!")
+    }
+    return checkstyleVersion
+}
+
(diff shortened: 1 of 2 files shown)


romani:
There should be no term "fix".
It is just comparison tool of two report that are done on some branches.
so name them as it is - "checkstyle-checkstyle-master" vs "MEZk-checkstyle-i3333".

checkstyle-tester/diff.groovy
```diff
@@ -0,0 +1,232 @@
...
+
+import static java.lang.System.err
+
+def cli = new CliBuilder(usage:'groovy diff.groovy [options]')
+cli.with {
+    r(longOpt: 'localGitRepo', args: 1, argName: 'localGitRepo', 'Path to local git repository')
+    b(longOpt: 'patchBranch', args: 1, argName: 'patchBranch', 'Name of the patch branch in local git repository')
+    c(longOpt: 'checkstyleCfg', args: 1, argName: 'checkstyleCfg', 'Path to checkstyle config file')
+    p(longOpt: 'projectsToTestOn', args: 1, argName: 'projectsToTestOn', 'Path to file which contains projects to test on')
+}
```
romani: we need to have:
path to repo - "r"
base branch name ("master" is default) - "b"
patched/changed branch name - "p"
checktyle config - "c"
path to properties file with projects - "l" (list of projects)

checkstyle-tester/diff.groovy
```diff
@@ -0,0 +1,232 @@
...
+def options = cli.parse(args)
```
romani: please make a "main" method , to show no magic from where `args` appear

checkstyle-tester/diff.groovy
```diff
@@ -0,0 +1,232 @@
...
+        throw new IllegalStateException("Error: git repository ${localGitRepo.getPath()} has unstaged changes!")
+    }
+
+    def patchBranch = options.patchBranch
+    def projectsToTestOn = options.projectsToTestOn
+    def checkstyleCfg = options.checkstyleCfg
+
+    def reportsDir = 'reports'
+    def reportsBeforeFixDir = "$reportsDir/before_fix"
+    def reportsAfterFixDir = "$reportsDir/after_fix"
```
romani: There should be no term "fix".
It is just comparison tool of two report that are done on some branches.
so name them as it is - "checkstyle-checkstyle-master" vs "MEZk-checkstyle-i3333".
In terms of this script: localGitRepo v patchBranch ..... master vs i3333

checkstyle-tester/diff.groovy
```diff
@@ -0,0 +1,232 @@
...
+    if (new File(tmpReportsDir).exists()) {
+        deleteDir(tmpReportsDir)
+    }
+
+    generateCheckstyleReport(localGitRepo, 'master', checkstyleCfg, projectsToTestOn, tmpReportsBeforeFixDir)
+    generateCheckstyleReport(localGitRepo, patchBranch, checkstyleCfg, projectsToTestOn, tmpReportsAfterFixDir)
+    deleteDir(reportsDir)
+    moveDir(tmpReportsDir, reportsDir)
+    generateDiffReport(reportsDir, reportsBeforeFixDir, reportsAfterFixDir, checkstyleCfg)
+    createSummaryIndexHtml(diffDir)
```
romani: generateSummaryIndexHtml

checkstyle-tester/diff.groovy
```diff
@@ -0,0 +1,232 @@
...
+    if (gitStatusOutput.contains("nothing to commit, working directory clean")) {
+        hasUnstagedChanges = false
+    }
+    return hasUnstagedChanges
+}
+
+def generateCheckstyleReport(localGitRepo, branch, checkstyleCfg, projectsToTestOn, destDir) {
+    println "Installing Checkstyle artifact ($branch) into local Maven repository ..."
+    executeCmd("git checkout $branch", localGitRepo)
+    executeCmd("mvn -Passembly clean install", localGitRepo)
```
romani: -Pno-validation
I just did renaming in main repo.

MEZk:
@romani 
>There should be no term "fix

Done.

MEZk:
@romani 
>we need to have:
path to repo - "r"
base branch name ("master" is default) - "b"
patched/changed branch name - "p"
checktyle config - "c"
path to properties file with projects - "l" (list of projects)

Done.

>please make a "main" method , to show no magic from where args appear

Done.

>There should be no term "fix".

Done.

>generateSummaryIndexHtml

Done.

>-Pno-validation
I just did renaming in main repo.

Done

romani:
Im ok with code,  @rnveach please review and merge

MEZk:
@romani 

Should I update README.md with information on how to use launch.groovy and diff.groovy as a separate commit?

romani:
Either is fine

checkstyle-tester/diff.groovy
```diff
@@ -0,0 +1,243 @@
...
+
+    if (areValidCliArgs(options)) {
+        def localGitRepo = new File(options.localGitRepo)
+        if (hasUnstagedChanges(localGitRepo)) {
+            throw new IllegalStateException("Error: git repository ${localGitRepo.getPath()} has unstaged changes!")
+        }
+
+        def baseBranch = options.baseBranch
+        if (!baseBranch) {
+            baseBranch = 'master'
```
rnveach: default base is local master? Shouldn't we rely more on remote master as that is the true base?
Note: if this is changed to remote, we will have to put in a fetch to make sure it is always latest.

checkstyle-tester/diff.groovy
```diff
@@ -0,0 +1,243 @@
...
+        hasUnstagedChanges = false
+    }
+    return hasUnstagedChanges
+}
+
+def generateCheckstyleReport(localGitRepo, branch, checkstyleCfg, projectsToTestOn, destDir) {
+    println "Installing Checkstyle artifact ($branch) into local Maven repository ..."
+    executeCmd("git checkout $branch", localGitRepo)
+    executeCmd("mvn -Pno-validations clean install", localGitRepo)
+    executeCmd("groovy launch.groovy $projectsToTestOn $checkstyleCfg")
```
rnveach: Note: No passing and verifying that checkstyle version POM matches tester's checkstyle version.
If these don't match, reports will falsely show no changes when there could be changes.
https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/launch_diff.sh#L110-L121
This is critical we add this imo.

checkstyle-tester/diff.groovy
```diff
@@ -0,0 +1,243 @@
...
+    println "Moving Checkstyle report into $destDir ..."
+    moveDir("reports", destDir)
```
rnveach: Just confirming since you move directories around, did you confirm refs still point to valid location in final report for diff-tool?
For my tool I had to modify them.
https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/launch_diff.sh#L271

checkstyle-tester/diff.groovy
```diff
@@ -0,0 +1,243 @@
...
+}
+
+def generateDiffReport(reportsDir, masterReportsDir, patchReportsDir, checkstyleCfg) {
+    def diffToolDir = Paths.get("").toAbsolutePath()
+        .getParent()
+        .resolve("patch-diff-report-tool")
+        .toFile()
+    executeCmd("mvn clean package", diffToolDir)
```
rnveach: do we need to create diff report tool everytime? Can we skip if it exists already?
Would be cool to skip if JAR exists and fetch produced no new changes, but we can skip complicated for now,

checkstyle-tester/diff.groovy
```diff
@@ -0,0 +1,243 @@
...
+    println 'Starting diff report generation ...'
+    Paths.get(masterReportsDir).toFile().eachFile {
+        fileObj ->
+            if (fileObj.isDirectory()) {
+                def projectName = fileObj.getName()
+                def patchReportDir = new File("$patchReportsDir/$projectName")
+                if (patchReportDir.exists()) {
+                    def baseReport = "$masterReportsDir/$projectName/checkstyle-result.xml"
+                    def patchReport = "$patchReportsDir/$projectName/checkstyle-result.xml"
+                    def refFiles = "src/main/java"
```
rnveach: this is correct ref even though we moved files around?

checkstyle-tester/diff.groovy
```diff
@@ -0,0 +1,243 @@
...
+                    def outputDir = "$reportsDir/diff/$projectName"
```
rnveach: we don't need to create directory on first run?

checkstyle-tester/diff.groovy
romani: I am not agree on this request.
we should take that user give us and point to

checkstyle-tester/diff.groovy
romani: removal of "clean" will help.

checkstyle-tester/diff.groovy
rnveach: @romani Users can modify their local branches and put in any commits they want and we won't be able to tell the difference. Reports don't mention commit hashes they are working with. They are also not required to keep it up to date. With remote master (`origin/master`) and fetching, we know we are always dealing with the latest changes and what everyone else is using.

romani: I am ok print hashes and timestamps in reports, but still think that forced fetch is not good

MEZk: @rnveach 
I do not agree with your idea to fetch remote changes. Sometimes I work on Checkstyle in areas where I do not have Internet access for a couple of days, still I want to generate diff reports to make sure that my changes are correct.

checkstyle-tester/diff.groovy
MEZk: @rnveach
Diff dir will be created by diff tool (--output <diff_dir>)

checkstyle-tester/diff.groovy
romani: @rnveach , please open separate issue to discuss this separately, update it separately.
This point should not block merge of this PR.

rnveach: @romani I am fine with printing the hashes.
If I am firm on something, I will let you know. :)

romani: @MEZk , please create issue on printing hashes, we will do this separately, this PR become big.
Lets merge it while it is manageable state.

checkstyle-tester/diff.groovy
MEZk: Fixed for launch.groovy script in separate PR (https://github.com/checkstyle/contribution/pull/140)

checkstyle-tester/diff.groovy
MEZk: @romani 
Done. https://github.com/checkstyle/contribution/issues/141

checkstyle-tester/diff.groovy
MEZk: @rnveach 
#140 fixes the problem. I checked references in diff reports.

checkstyle-tester/diff.groovy
MEZk: @rnveach 
I will update README.md and clearly state all the process in separate PR.

MEZk:
@romani @rnveach 
All points were addressed.

checkstyle-tester/diff.groovy
rnveach: @MEZk How will change in `launch.groovy` help here as the move directory command is in this script? If you are moving twice, don't you need replace the refs twice?

MEZk: @rnveach 
Only references in checkstyle-result.xml need to be changed to point to sources inside repositories directory.

checkstyle-tester/diff.groovy
```diff
@@ -0,0 +1,262 @@
...
+    def options = cli.parse(args)
+
+    if (areValidCliArgs(options)) {
+        def localGitRepo = new File(options.localGitRepo)
+        if (hasUnstagedChanges(localGitRepo)) {
+            throw new IllegalStateException("Error: git repository ${localGitRepo.getPath()} has unstaged changes!")
+        }
+        def testerCheckstyleVersion = getCheckstyleVersionFromPomXml('./pom.xml', 'checkstyle.version')
+        def checkstyleVersionInLocalRepo = getCheckstyleVersionFromPomXml("$localGitRepo/pom.xml", 'version')
+        if (testerCheckstyleVersion != checkstyleVersionInLocalRepo) {
```
rnveach: The full point is not addressed.
You are checking pom in checkstyle git, but not for master and patch branches. You are checking it for some random initial branch. Since no checkouts have been called, we can't tell what we are running this check on.

https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/launch_diff.sh#L118-L129
`launch_diff` checks the POM for both branches. It is in `launch` function which is called twice.
Master: https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/launch_diff.sh#L363
Patch: https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/launch_diff.sh#L398

PR #140, opened 2016-12-14
Issue #51: Fix paths in checkstyle-result.xml

MEZk:


romani:
It is hard to measure this PR from reading it. I am ok.
@rnveach , please review and merge.

MEZk:
@rnveach 
This fix is for checkstyle-result.xml. It replaces 'src/main/java' in all paths inside checkstyle-result.xml with 'repositories' since all sources will be located in checkstyle-tester/repositories after launch.groovy is being executed.

checkstyle-tester/launch.groovy
```diff
@@ -207,13 +206,20 @@ def runMavenExecution(srcDir, excludes, checkstyleConfig) {
...
     println 'linking report to index.html'
     new File("$siteDir/index.html").renameTo "$siteDir/_index.html"
     Files.createLink(Paths.get("$siteDir/index.html"), Paths.get("$siteDir/checkstyle.html"))
 
     removeNonReferencedXrefFiles(siteDir)
     removeEmptyDirectories(new File("$siteDir/xref"))
+
+    new AntBuilder().replace(
+        file: "$targetDir/checkstyle-result.xml",
+        token: "src/main/java",
```
rnveach: I don't know this replace method, but what happens if a project we are regressing on has a directory called `src/main/java`?
Lets take this as example: `/home/rnveach/checkstyle-tester/src/main/java/my_project_abcdef/src/main/java/...`
Won't it replace it to be `/home/rnveach/checkstyle-tester/repositories/my_project_abcdef/repositories/...` instead of `/home/rnveach/checkstyle-tester/repositories/my_project_abcdef/src/main/java/...`? The second should be the correct final form, not the first.
If we are to do replaces like this, we should probably be using the fully qualified path.

MEZk: @rnveach 
Fixed.
Now it replaces all occurrences of 'checkstyle-tester/src/main/java' with 'checkstyle-tester/repositories'

MEZk:
@romani 
Take a look at @rnveach commend and my fix. Are we ok to merge the PR since @rnveach is on a vocation?

rnveach:
@MEZk vacation is not until next week :)

PR #142, opened 2016-12-16
doc: [checkstyle-tester] Update README.md to explain how to use new scripts written in Groovy

MEZk:
@romani 
Updated README.md


romani:
thanks a lot.
if smth is missed .... could be added in next PR.

Issue #152, opened 2017-01-21
Turn tester failsOnError into a property


rnveach:
https://github.com/checkstyle/contribution/blob/3ffb678b666e3e26c4227ca41d5a35237be343cd/checkstyle-tester/pom.xml#L59
This line needs to be turned into a property as a default value of `false` damages no exception testing.

In no exception testing, we want an exception to result in a build failure.
In a regression testing, we don't want an exception to result in a build failure, but be listed as a type of violation in the reports for differences.

`groovy` and `launch_diff` need to be changed to specify and set the property to `false`.
Default property in POM should be `true`.

romani:
https://maven.apache.org/plugins/maven-checkstyle-plugin/check-mojo.html#failsOnError

doe not have user property, so we should move it to local property of our pom file.

rnveach:
Since groovy's diff is 2 separate programs, to enable this feature in it, we would have to have one send the other an optional argument.
Maybe we could use [CliBuilder](http://docs.groovy-lang.org/2.4.7/html/gapi/groovy/util/CliBuilder.html)

romani:
fix is merged

rnveach:
@romani Fix wasn't applied to groovy program.

MEZk:
@romani 
Assign. I will provide a fix.

romani:
@MEZk, discuss this with rnveach.

rnveach:
@MEZk go right ahead if you want to.

rnveach:
the 2 fixes close the issue

PR #156, opened 2017-01-21
Issue #152: turned tester failsOnError into property

rnveach:
Issue #152

added property and changed `launch_diff`.
Still need to change groovy which may take me some time since I never worked with it before.


MEZk:
@rnveach 
The same command is in launch.groovy
https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/launch.groovy#L205

I can add option in groovy script.

rnveach:
@MEZk yes, but it must get the value to turn it on from `diff.groovy` as this script alone is used for no-exception testing, which should always fail on an exception.

Issue #160, opened 2017-01-22
Groovy tester script should clean out working directory beforehand


rnveach:
Issue is seen https://github.com/checkstyle/checkstyle/pull/3737#issuecomment-274291107 and https://github.com/checkstyle/checkstyle/pull/3737#issuecomment-274294807.

Groovy isn't cleaning out the `src/main/java` before working with it.
The scripts should have complete control over this directory and should clean it out before adding any projects to it during regression.

MEZk:
@rnveach 
Do we need an empty file in src/main/java ?

rnveach:
@MEZk We may need it in Git just so the directory exists, but if scripts start running, only thing in there should be the project(s) assigned to be tested on.
If I am testing `Project A`, and the directory already has `Project Z` in it, then Z should be cleared out before A is populated in there and reported on. If we leave Z in there, then it will appear in the reports when all we are expecting is `Project A`.

This answer your question? If not, please explain more.

romani:
We did not do removal content of this folder to let user rerun testing without recline from web that is long for openjdk for example

rnveach:
> rerun testing

@romani These scripts do not support resuming.
[In launch.groovy, if multiple projects are set, we do each one separately clearing out the directory](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/launch.groovy#L64). If `Project B` failed somewhere, rerunning would still start at `Project A` not B. In the scenario of this issue, this is what happened and `Project B`s excludes were not included in the run of A, which resulted in exceptions and a build failure.
[In diff.grovy, we always start with base run](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/diff.groovy#L49). If patch run fails, we cannot restart there.
[`launch_diff.sh` supports some skips at major intersections](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/launch_diff.sh#L67-L98), but it will never resume on a specific project.

It is not apparent to the user that a blank run is actually a resume run. If we want to fully support resuming, it should be with some specific option in the command line.

> decline from web that is long

I assume you are talking about travis and CI? Unless we cache the directory, travis and other CIs won't do a resume run because they always start with a blank slate, right?

romani:
Thanks a lot for explanation, it was my problem of mixing diff tool and just report tool.
No resume is required.

CI are running on clean environment.

romani:
Fix is merged

PR #161, opened 2017-01-22
Issue #160: Fix problem with diff report generation due to non-empty src directory

MEZk:
#160 

@rnveach 
Please, take a look at the changes.


MEZk:
@romani 
Please accept PR.

Issue #176, opened 2017-02-06
patch-diff-report-tool: Add tests and code coverage


rnveach:
We should start adding tests and coverage to make sure utility is stable as we are using it more and more.

romani:
please do not be crazy with over testing.

If that is required , it is already sign that tool need to be written in java and have complete set of validations.
We did this tool on groovy only to be a bit more user friendly (not a shell) and keep simple (script like).
In all my experience with groovy, any investment in complicated groovy will result in high cost of support. As you grow script it should become multi-file , so not a script any more so need to be compiled , so..... become the same as java.

As I told you in other issue, it is enough to keep in CI full run with recheck of expected output.

rnveach:
@romani 
> need to be written in java .... We did this tool on groovy only 

I think you are confusing what tool I am talking about here.
[patch-diff-report-tool](https://github.com/checkstyle/contribution/tree/master/patch-diff-report-tool) is already in Java, not groovy. I am not talking about making tests for groovy scripts here.

romani:
yes,  my bad.
tests for java diff tool is ok.

PR #178, opened 2017-02-07
Issue #176: started adding tests for patch-diff-report-tool

rnveach:
Issue #176 

I will be splitting this issue up some.
The start of adding tests. Code coverage is not introduced yet.
I only removed dead code from the `main` folder that I was sure couldn't be executed.
`run` folder has empty files as they are required to exist but aren't required to be populated.
All XML files started out valid, but to appease some coverage I had to add invalid tags which don't interfere as the DTD is not attached to validate them.

Current line coverage is 93% total. Branch coverage is 79.5%.


rnveach:
> `ComparisonFailure: expected:<...	<td><a href = "xref[\]File1.html#L5">5</a></td>...> but was:<...	<td><a href = "xref[/]File1.html#L5">5</a></td>`

Forgot about file path separator. I'll need to provide a fix for this.

rnveach:
Tests were fixed, CI is passing.

PR #179, opened 2017-02-07
Pull #179: transformed xref windows url to linux

rnveach:
This is the fix for PR #178 .
File path in the report's href was the only area where it was different depending on the OS being used.

> `ComparisonFailure: expected:<...	<td><a href = "xref[\]File1.html#L5">5</a></td>...> but was:<...	<td><a href = "xref[/]File1.html#L5">5</a></td>`


rnveach:
Closing PR.

As discussed in chat, report consistency between different OSes doesn't matter. Test should be updated to fix static test file itself to avoid conflicts.

Issue #206, opened 2017-03-28
releasenotes-builder: missed new feature


romani:
issue https://github.com/checkstyle/checkstyle/issues/3437
one more : https://github.com/checkstyle/checkstyle/issues/3892

has label "new feature" but it is not present in releasenotes after 7.6, that forced me to create 7.6.1 release instead of 7.7 .

Release notes skip information for update for all our plugins. That might result in wave of problems.

TODO: investigate a reason of missed "new feature" group in release notes of 7.6.1 - checkstyle.sourceforge.net/releasenotes.html#Release_7.6.1

link to all closed "new feature" issues - https://github.com/checkstyle/checkstyle/issues?q=is%3Aissue+label%3A%22new+feature%22+is%3Aclosed+sort%3Aupdated-desc

romani:
one more problem - https://github.com/checkstyle/checkstyle/issues/3329 does not have label  and build is not failed

romani:
problem is even bigger:

```
$git log --pretty=format:"%s" checkstyle-7.6...checkstyle-7.6.1 | grep -vE "^config:|minor:|infra:|spelling:|maven-release|^doc:" > 1.log

2.log is created by copy-paste of web content

$ cut -d " " -f2 1.log | cut -c2-5 | sort | uniq > 1_numbers.log

$ cut -d "#" -f2 2.log | grep -E "^[0-9]" | sort | uniq > 2_numbers.log

$ diff -u 1_numbers.log 2_numbers.log 
--- 1_numbers.log   2017-03-28 11:49:10.040920921 -0700
+++ 2_numbers.log   2017-03-28 11:49:14.329127713 -0700
@@ -2,10 +2,7 @@
 3172
 3329
 3437
-3491
 3612
-3648
-3685
 3700
 3731
 3755
@@ -14,38 +11,20 @@
 3843
 3848
 3850
+3868
 3869
 3883
 3888
 3892
 3896
-3897
-3900
 3904
 3906
 3908
-3910
-3929
-3943
-3954
-3958
-3961
 3962
-3965
-3970
-3976
-3985
-3986
 3989
-3991
-3999
-4003
-4014
 4030
-4042
 4044
 4045
 4055
 4081
 4093
-4095

$ diff -u 1_numbers.log 2_numbers.log | grep -E "^(\+|-)[0-9]"
-3491
-3648
-3685
+3868
-3897
-3900
-3910
-3929
-3943
-3954
-3958
-3961
-3965
-3970
[...]

PR #208, opened 2017-03-29
Issue #206: Fix incorrect filter condition for binary search

MEZk:
#208 

@romani 
> This looks like the issue. Binary search returns >= 0 if it exists in array and < 0 on where it should be inserted into array if it doesn't exist. It doesn't guarantee a result of -1 all the time.
Condition should be checking >= 0 instead of != -1.

Yes, that was the problem.

> issue checkstyle/checkstyle#3437
one more : checkstyle/checkstyle#3892
has label "new feature" but it is not present in releasenotes after 7.6, that forced me to create 7.6.1 release instead of 7.7 .
TODO: investigate a reason of missed "new feature" group in release notes of 7.6.1 - checkstyle.sourceforge.net/releasenotes.html#Release_7.6.1

Fixed.

> one more problem - checkstyle/checkstyle#3329 does not have label and build is not failed

Fixed.

> link to all closed "new feature" issues - https://github.com/checkstyle/checkstyle/issues?q=is%3Aissue+label%3A%22new+feature%22+is%3Aclosed+sort%3Aupdated-desc

Here is the [xdoc.xml](https://github.com/checkstyle/contribution/files/880123/xdoc.txt) which was generated starting from tag ```checkstyle-6.12```. As you can see there are 48 issues in ```New``` group.

Program args to check:
```-localRepoPath /checkstyle -startRef checkstyle-6.12 -githubAuthToken *********************************** -releaseNumber 8 -generateXdoc```

I do not know why TeamCity is not happy this time.

### DIFF
--- a/releasenotes-builder/src/main/java/com/github/checkstyle/NotesBuilder.java
+++ b/releasenotes-builder/src/main/java/com/github/checkstyle/NotesBuilder.java
@@ -306,7 +306,7 @@ public final class NotesBuilder {
     private static String getIssueLabelFrom(GHIssue issue) throws IOException {
         final Collection<GHLabel> issueLabels = issue.getLabels();
         final Optional<GHLabel> label = issueLabels.stream()
-            .filter(input -> Arrays.binarySearch(Constants.ISSUE_LABELS, input.getName()) != -1)
+            .filter(input -> Arrays.binarySearch(Constants.ISSUE_LABELS, input.getName()) >= 0)
             .findFirst();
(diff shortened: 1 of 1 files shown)


rnveach:
> I do not know why TeamCity is not happy this time.

TC problems are because they upgraded, and the upgrade is not working correctly and it is breaking everything.

MEZk:
@rnveach 
> TC problems are because they upgraded, and the upgrade is not working correctly and it is breaking everything.

It's a pity. This is not the first time (

Issue #213, opened 2017-04-12
checkstyle-tester: support regression on new checks/tokens


rnveach:
This has come up twice in PRs where regression is requested. The latest is at https://github.com/checkstyle/checkstyle/pull/4149#issuecomment-293634837 .

When a new token is introduced into an existing check, there is no way to run `diff.groovy`.
Groovy requires a before image (master) and an after image (PR branch). When specifying a brand new token alone, there is nothing to run it against in the before image as the token doesn't exist. Attempting to run a configuration with a token that doesn't exist will get something similar to the following error:
> Caused by: com.puppycrawl.tools.checkstyle.api.CheckstyleException: Token "METHOD_REF" was not found in Acceptable tokens list in check com.puppycrawl.tools.checkstyle.checks.whitespace.SeparatorWrapCheck

Users are forced to run `launch.groovy` which doesn't automatically generate a main index page, which allows us to instantly see which projects have differences/violations.

We should add support for these types of scenarios in `diff.groovy` and `patch-diff-report-tool`.
For this option to be activated, the user should only need to specify the patch branch and the patch config.
We should use a special option in `diff.groovy` to specify this type of run like `--patchOnly`.

It should act like the before image run produced no violations.
This type of option will reduce the regression run by half as it is only doing the PR run and not the master run.

MEZk:
@rnveach 
diff.groovy supports xml configuration for the branch before changes (master) and PR branch: baseConfig and patchConfig. So, what is the problem with it? You need to pass to configs to diff.groovy: one without a new token and one with.

rnveach:
> there is no way to run launch.groovy.

I'm sorry, I meant `diff.groovy`.
I will change the post.

> You need to pass to configs to diff.groovy: one without a new token and one with.

What do you suggest for before image when token is brand new and is not supported in master and we just want violations on just that token?
The only way I see is to apply all acceptable tokens to the check in both and hope that regression of the tokens don't intersect one another. [We have one check, `ParenPad`, which had this happen](https://github.com/checkstyle/checkstyle/issues/4175). One token was being picked up in master when another was set, and regression always showed 0 differences until we ran regression on single tokens.
So adding all tokens will just give us the regression of all those tokens, not regression on the individual isolated tokens.

> What do you mean? I do not understand the idea. Please, explain.

[...]

Issue #182, opened 2017-02-08
Refactoring of diff.groovy


MEZk:
1) generateDiffReport and generateCheckstyleReport now receive too many parameters. Parameters should be represented as an objects of the specific class or the methods need to be refactored so as not to have so many parameters.

romani:
to my mind is it fine to have that much arguments for now.

> Parameters should be represented as an objects

please do not over complicate the problem. All arguments could be refactored to single structure/object as wrapper.

MEZk:
should be done after #213

PR #215, opened 2017-04-20
Issue #213: checkstyle-tester: support regression on new checks/tokens

MEZk:
issue #213 

1) Added ability to generate Checkstyle reports for patch branch only with diff.groovy
2) Extended README.md
3) diff.groovy does not have ```-i``` command line option (only launch groovy has), so it was removed from Travis CI yml.

Further refactoring and performance optimization of diff.groovy will be performed in the scope of https://github.com/checkstyle/contribution/issues/182.

diff.groovy should be renamed to reports.groovy or something similar in separate issue.

@rnveach
Maybe we need to have the manual for launch.groovy in a separate file. Do contributors need to know about it from the main REAMDE.md of checkstyle-tester?

### DIFF
--- a/checkstyle-tester/README.md
+++ b/checkstyle-tester/README.md
@@ -1,109 +1,16 @@
 # CHECKSTYLE-TESTER
 
 checkstyle-tester is a tool for Checkstyle report generation over very [basic set of java projects](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/projects-to-test-on.properties).
-It consists of two Groovy scripts: launch.groovy and diff.groovy. Thus, in order to use the tool make sure you have the Groovy runtime installed on your developer machine (min required version is 2.4.8).
+It consists of one Groovy script: diff.groovy. Thus, in order to use the tool make sure you have the Groovy runtime installed on your developer machine (min required version is 2.4.8).
 
 Content:
 
-- [Checkstyle report generation](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/README.md#launchgroovy-checkstyle-report-generation)
 - [Diff report generation](https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#diffgroovy-diff-report-generation)
 - [Deploy Report](https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#deploy-report)
 
-
-## [launch.groovy] Checkstyle report generation
-
-**launch.groovy** is a script which allows you to generate Checkstyle report over target projects. It invokes Maven Checkstyle plugin. In order to use the script you should run the following command in your command line:
-
-```
-groovy launch.groovy --listOfProjects projects-to-test-on.properties --config my_check.xml
-```
-
-or with short command line arguments names:
-
-```
-groovy launch.groovy -l projects-to-test-on.properties -c my_check.xml
-```
-
-If you want to force Maven Checkstyle Plugin to ignore exceptions:
-
-```
-groovy launch.groovy --listOfProjects projects-to-test-on.properties --config my_check.xml --ignoreExceptions
-```
-
-or with short command line arguments names:
-
-```
-groovy launch.groovy -l projects-to-test-on.properties -c my_check.xml -i
-```
-
-The script receives the following command line arguments:
-
-**listOfProjects** (l) - path to the file which contains the projects which sources will be analyzed by Checkstyle during report generation (required);
-
-**config** (c) - path to the file with Checkstyle configuration (required). The default config should be changed in order to be appropriate for your use purposes;
-
-**ignoreExceptions** (i) - whether Checkstyle Maven Plugin should ignore exceptions (optional, default is false).
-
-**ignoreExcludes** (g) - whether to ignore excludes specified in the list of projects (optional, default is false).
-
-When the script finishes its work the following directory structure will be created in the root of cehckstyle-tester directory:
-
-*/repositories* - directory with downloaded projects sources which are specified in projects-to-test-on.properties;
-
-*/reports* - directory with Checkstyle reports. 
-
-*reports/{repository_name}/site* - directory with Checkstyle report generated for the specific project (repository);
-
-You will find *index.html* file in /reports/{repository_name}/site directory. The file represents the main page of the report.
-
-**First run** should be executed with all rules enabled to make sure that new check does not fail. 
-You may see failure of `TreeWalker` but as long as it is applied for no-compilable sources (test, resources) you don't need to worry about it.  **Second run** shall be done to prove that check is working correctly and for this one select the most accurate repositories.
-
-Please use `-Dcheckstyle.consoleOutput=true` option if you need to see all Checkstyle errors in console, but expect it to print very and very much lines of terminal logs in this case. Launch command in this case will be:
-
-```
-groovy launch.groovy projects-to-test-on.properties my_check.xml -Dcheckstyle.consoleOutput=true
-```
-
-**Attention:** this project by default uses the latest SNAPSHOT version of [Checkstyle](https://github.com/checkstyle/contribution/search?utf8=%E2%9C%93&q=path%3Acheckstyle-tester+filename%3Apom.xml+%22checkstyle.version%22&type=) and the latest released version of [sevntu.checkstyle](https://github.com/checkstyle/contribution/search?utf8=%E2%9C%93&q=path%3Acheckstyle-tester+filename%3Apom.xml+%22sevntu.checkstyle.version%22&type=).
-If you you need to use custom (snapshot) versions please update pom.xml to reference that versions ([checkstyle version](https://github.com/checkstyle/contribution/blob/35d35dfcc48e2022403231e41aac8bf96126acc9/checkstyle-tester/pom.xml#L15), [sevntu.checkstyle version](https://github.com/checkstyle/contribution/blob/35d35dfcc48e2022403231e41aac8bf96126acc9/checkstyle-tester/pom.xml#L16)), and please make sure that custom(newly generated) versions are located in your local maven repo 
-
-```
-ls  ~/.m2/repository/com/puppycrawl/tools/checkstyle/
-ls  ~/.m2/repository/com/github/sevntu/checkstyle/sevntu-checkstyle-maven-plugin
-```
-
-To build SNAPSHOT version of `checkstyle` please run in its folder (local git repository):
-
-```
-mvn clean install -DskipTests -Dcobertura.skip=true -Dfindbugs.skip=true -Dpmd.skip=true
-```
-
-**Attention:** 
-Make sure that the versions of Checkstyle and SevNTU Checkstyle artifacts, which are located in your local Maven repository, corresponds to the versions specified in pom.xml,
-1) otherwise there will be the exception, because maven-checkstyle-plugin will not find the required artifact in the local Maven repository;
-2) you will not see any new/different violations from your changes, especially if you make changes to version X with changed logic and pom.xml specifies version Y. 
-You will not get an exception message and unless you are expecting a specific difference, you may just assume your report is correct when it really isn't.
-launch.groovy has no way to protect against this as branch with change isn't specified.
-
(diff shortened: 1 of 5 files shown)


rnveach:
> diff.groovy should be renamed to reports.groovy 

Any reason why? I think it makes more sense as it does more than create the reports, it generates the differences.

> Do contributors need to know about it from the main REAMDE.md of checkstyle-tester?

Would it make more sense for each utility to have it's own readme file?
If not, I am fine with it staying in the same file as we can link directly to diff.groovy.

checkstyle-tester/diff.groovy
```diff
@@ -161,6 +135,20 @@ def hasUnstagedChanges(gitRepo) {
     return hasUnstagedChanges
 }
 
+def runPlainCheckstyleReportsGeneration(cfg) {
+    generateCheckstyleReport(cfg.localGitRepo, cfg.patchBranch, cfg.patchCheckstyleCfg, cfg.listOfProjects, cfg.reportsDir)
+    generateSummaryIndexHtml(cfg.reportsDir)
```
rnveach: @MEZk In the end, I only want to look at diff reports and not the summary report. :)
I think our report is alot easier to view.

If we made `patch-diff-report-tool` support no base config or base reports, we wouldn't need this extra parsing or different code for the summary index. Then all reports would look the same and it would be easier to make more additions to the summary, if we decide to.

Does this sound like a better idea?
@romani What do you think?

MEZk:
@rnveach 

> Would it make more sense for each utility to have it's own readme file?
If not, I am fine with it staying in the same file as we can link directly to diff.groovy.

I guess, yes. Contributor who wants to generate diff reports should not know about launch.groovy. However, those who want to fix the issues of launch.groovy and checkstyle-tester project itself, should. That is why I think it is better to have separate readme files.

checkstyle-tester/diff.groovy
MEZk: @rnveach 
> In the end, I only want to look at diff reports and not the summary report. :)
I think our report is alot easier to view.

If I generate reports for more than one project, I think that having summary page makes it easier to review the reports which are uploaded to github.io.

> If we made patch-diff-report-tool support no base config or base reports, we wouldn't need this extra parsing or different code for the summary index. Then all reports would look the same and it would be easier to make more additions to the summary, if we decide to.

Agree.

rnveach: >  I think that having summary page 

I don't know how I keep mixing things up. 
When I said summary report, I meant the normal `launch.groovy` report. I'm sorry again.

http://rveach.no-ip.org/checkstyle/regression/reports/20/checkstyle/index.html I meant this report.

rnveach: PR for patch-diff-report-util started at https://github.com/checkstyle/contribution/pull/216

romani:
if we so care about new contributors and we do not want to support two separate scripts

we need to rename `diff.groovy` to `cs-tester.groovy` and make separate modes for diff and single launch.

rnveach:
@MEZk PR for `patch-diff-report` was merged. We can continue this PR based on new comments romani made in issue.
I'm not sure which issue this will be filed under yet.

MEZk:
@rnveach 
Updated.

@romani 
The build failed due to Codenarc's violaions. I added 2 def tokens as I still thing that we should use the Groovy's keyword. Please, merge https://github.com/checkstyle/contribution/pull/220 to let me restart the build and get the actual Codenarc's violations.

checkstyle-tester/README.md
```diff
@@ -1,138 +1,59 @@
...
 ```
 
 The script receives the following set of command line arguments:
 
 **localGitRepo** (r) - path to the local Checkstyle repository (required);
 
-**baseBranch** (b) - name of the base branch in local Checkstyle repository (optional, default is master branch);
+**baseBranch** (b) - name of the base branch in local Checkstyle repository (optional, if asent, then the tool will use only patchBranch in case patchOnly is set, otherwise baseBrach will be set to 'master');
+
+**patchOnly** (po) - whether to generate report only for the patch branch. If the option is set, then baseBranch, baseConfig and config should not be specified as the tool will finish the execution with the error. You must specify patchBranch and patchConfig if you enable this option. (optional, not specified by default)
```
rnveach: romani wants the option to be named `mode` with the values `diff` (default) or `single`.
https://github.com/checkstyle/contribution/issues/213#issuecomment-297531342

I think we should explicitly tell users why they would use this option.
This is for any option that can't work on master/base branch, which are new properties, new tokens, or new modules.

checkstyle-tester/README.md
```diff
@@ -152,7 +73,7 @@ When the script finishes its work the following directory structure will be crea
 
 */reports/diff* - directory with diff reports;
 
-*reports/baseBranch* - directory with Checkstyle reports which are generated with Checkstyle base version (based on specified base branch);
+*reports/baseBranch* - directory with Checkstyle reports which are generated with Checkstyle base version (based on specified base branch. If patchOnly is enabled then the diractory will not be created.);
```
rnveach: `diractory` -> `directory`

checkstyle-tester/diff.groovy
```diff
@@ -103,19 +109,20 @@ def areValidCliOptions(cliOptions) {
...
-def isValidCheckstyleConfigsCombination(config, baseConfig, patchConfig) {
+def isValidCheckstyleConfigsCombination(config, baseConfig, patchConfig, patchOnly) {
     def valid = true
-    if (config && patchConfig
+    if (baseConfig && !patchConfig
+            || config && patchConfig
             || config && baseConfig
             || config && patchConfig && baseConfig) {
-        err.println "Error: you should specify either \'config\' or \'baseConfig\' and \'patchConfig\'!"
+        err.println "Error: you should specify either \'config\' or \'baseConfig\' and \'patchConfig\', or \'patchConfig\' only!"
```
rnveach: I think there should be a comma before first 'or' too to more clearly make the separation from the 3 items.

checkstyle-tester/diff.groovy
```diff
@@ -103,19 +109,20 @@ def areValidCliOptions(cliOptions) {
...
 }
 
-def isValidCheckstyleConfigsCombination(config, baseConfig, patchConfig) {
+def isValidCheckstyleConfigsCombination(config, baseConfig, patchConfig, patchOnly) {
     def valid = true
-    if (config && patchConfig
+    if (baseConfig && !patchConfig
+            || config && patchConfig
             || config && baseConfig
             || config && patchConfig && baseConfig) {
```
rnveach: unrelated to changes but how can this or statement be hit while the others aren't?
Whenever this line is `true` previous 2 lines are `true` as well.

checkstyle-tester/diff.groovy
```diff
@@ -66,11 +70,12 @@ def getCliOptions(args) {
         r(longOpt: 'localGitRepo', args: 1, required: true, argName: 'path', 'Path to local git repository (required)')
         b(longOpt: 'baseBranch', args: 1, required: false, argName: 'branch_name', 'Base branch name. Default is master (optional, default is master)')
         p(longOpt: 'patchBranch', args: 1, required: true, argName: 'branch_name', 'Name of the patch branch in local git repository (required)')
-        bc(longOpt: 'baseConfig', args: 1, required: false, argName: 'path', 'Path to the base checkstyle config file (required if patchConfig is specified)')
+        bc(longOpt: 'baseConfig', args: 1, required: false, argName: 'path', 'Path to the base checkstyle config file (optional, optional, if asent then the tool will use only patchBranch in case patchOnly is set, otherwise baseBrach will be set to \'master\')')
```
rnveach: `asent` -> `absent`
`baseBrach` -> `baseBranch`

checkstyle-tester/README.md
MEZk: Done

checkstyle-tester/README.md
MEZk: Done

checkstyle-tester/diff.groovy
MEZk: Fixed

checkstyle-tester/diff.groovy
MEZk: Removed unnecessary condition.

checkstyle-tester/diff.groovy
MEZk: Done.

checkstyle-tester/README.md
```diff
@@ -1,138 +1,61 @@
...
 ```
 
 The script receives the following set of command line arguments:
 
 **localGitRepo** (r) - path to the local Checkstyle repository (required);
 
-**baseBranch** (b) - name of the base branch in local Checkstyle repository (optional, default is master branch);
+**baseBranch** (b) - name of the base branch in local Checkstyle repository (optional, if asent, then the tool will use only patchBranch in case the tool mode is 'single', otherwise baseBrach will be set to 'master');
+
+**mode** (m) - The mode of the tool: 'single' or 'diff'. Set this option to 'single' if your patch branch contains changes for any option that can't work on master/base branch. For example, for new properties, new tokens, or new modules.
```
rnveach: I don't see a mention if this is required or optional.
We should mention that `diff` is the default?

After the sentence "For example, ...", please add a sentence that says something like "For all other changes, 'diff' mode should be the preferred mode used." so it is clear.

MEZk: > I don't see a mention if this is required or optional.
We should mention that diff is the default?

It was mentioned at the end of the property description. I moved it at the start of the option description.

> After the sentence "For example, ...", please add a sentence that says something like "For all other changes, 'diff' mode should be the preferred mode used." so it is clear.

Done.

rnveach:
@MEZk #220 was merged. Please rebase and make CI green.
@romani Please do final review.

MEZk:
@rnveach 
Rebased.

@romani 
I still think that we should use the ```def``` keyword.

romani:
> I still think that we should use the def keyword.

lets vote. I presume there 3 of us this project.
@rnveach , please share your position

My position is to do not use "def".

MEZk:
@romani 
In your previous comment you suggest rewriting the code in the way it will become more similar to Java. This will mean that we should also prohibit the usage of Groovy's closures and other language structures and make Groovy code exactly the same as Java code.

rnveach:
@romani I am in favor of the `def` keyword to make defining variables stand out, like defining the type in Java.

romani:
ok,  2 against 1 , @MEZk , please disable a rule with wide explanation.

MEZk:
@romani @rnveach 
According to [Groovy Official Style guide](http://groovy-lang.org/style-guide.html):
> I’ll finish on some words on when and how to use optional typing. Groovy lets you decide whether you use explicit strong typing, or when you use def. I’ve got a rather simple rule of thumb: whenever the code you’re writing is going to be used by others as a public API, you should always favor the use of strong typing, it helps making the contract stronger, avoids possible passed arguments type mistakes, gives better documentation, and also helps the IDE with code completion. Whenever the code is for your use only, like private methods, or when the IDE can easily infer the type, then you’re more free to decide when to type or not.

As we design the tool which will be used and supported by other contributors, maybe it is even better to use strong typing. What do you think?

romani:
my position is not use "def". But I am fine with "def" if most of maintainers want it.

MEZk:
@romani 
Ok, lets leave def for now. All in all, style guide is a personal opinion of the man who wrote it, we can take into account his opinion, but still write code as we believe will be better.

romani:
CI has to be Green.

MEZk:
@romani 
The rule 'ExplicitCallToEqualsMethod' should be disabled as well, as it makes code cryptic. In Groovy ```==``` is an overloaded operator which invokes ```equasl```, but it is weird for Java contributors. I disable the rule.

romani:
Ok to disable this rule.
http://codenarc.sourceforge.net/codenarc-rules-groovyism.html

romani:
@MEZk, what stop you from running codenarc on local ? To send update that will pass finally.

>"The line exceeds 120"

Please fix this in this PR, as you will change code again. We have 15 violations like this.

MEZk:
@romani 
CIs are fixed.

PR #216, opened 2017-04-23
Issue #213: added support to patch-diff-report for patch only run

rnveach:
Issue #213


patch-diff-report-tool/src/main/java/com/github/checkstyle/parser/EmptyXmlEventReader.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Override
+    public boolean hasNext() {
+        return false;
+    }
+
+    @Override
+    public XMLEvent peek() throws XMLStreamException {
+        return null;
```
MEZk: Why does the method return null? Can it throw NoSuchElementException or UnsupportedOperationException to avoid malformed usage ?

patch-diff-report-tool/src/main/java/com/github/checkstyle/parser/EmptyXmlEventReader.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Override
+    public String getElementText() throws XMLStreamException {
+        return null;
```
MEZk: same here

patch-diff-report-tool/src/main/java/com/github/checkstyle/parser/EmptyXmlEventReader.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Override
+    public XMLEvent nextTag() throws XMLStreamException {
+        return null;
```
MEZk: and here

patch-diff-report-tool/src/main/java/com/github/checkstyle/parser/EmptyXmlEventReader.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Override
+    public Object getProperty(String name) throws IllegalArgumentException {
+        return null;
```
MEZk: and here

patch-diff-report-tool/src/main/java/com/github/checkstyle/Main.java
```diff
@@ -52,8 +52,7 @@
     public static final String MSG_HELP = "This program creates symmetric difference "
             + "from two checkstyle-result.xml reports\n" + "generated for checkstyle build.\n"
             + "Command line arguments:\n"
-            + "\t--baseReportPath - path to the base checkstyle-result.xml, "
-            + "obligatory argument;\n"
+            + "\t--baseReportPath - path to the base checkstyle-result.xml\n"
```
MEZk: point that the option is optional

patch-diff-report-tool/src/main/java/com/github/checkstyle/parser/EmptyXmlEventReader.java
rnveach: For these types of methods, I looked at the javadoc and it did not mention throwing `NoSuchElementException`. So instead of hunting down it's intended use, I decided to stay with `null`.
http://docs.oracle.com/javaee/5/api/javax/xml/stream/XMLEventReader.html#peek()

NSE is not a runtime exception, so we would have to use `XMLStreamException` or a different runtime exception.
Only `hasNext` has code coverage so it doesn't really matter for our purposes right now what they do.

patch-diff-report-tool/src/main/java/com/github/checkstyle/parser/EmptyXmlEventReader.java
```diff
@@ -0,0 +1,71 @@
...
+
+import javax.xml.stream.XMLEventReader;
+import javax.xml.stream.XMLStreamException;
+import javax.xml.stream.events.XMLEvent;
+
+/**
+ * Simple {@link XMLEventReader} with no elements.
+ * @author Richard Veach
+ */
+public final class EmptyXmlEventReader implements XMLEventReader {
```
rnveach: > Do we need to introduce EmptyXmlEventReader to keep logic unchanged in parse ? 

That was my main reasoning, yes.
We still need a change in `StaxUtils.createReader` because it will get a NPE. We also pass the reader around, so who knows if we would have to make more places null safe.

patch-diff-report-tool/src/main/java/com/github/checkstyle/parser/EmptyXmlEventReader.java
rnveach: same answer as above.

patch-diff-report-tool/src/main/java/com/github/checkstyle/parser/EmptyXmlEventReader.java
rnveach: same answer as above.

patch-diff-report-tool/src/main/java/com/github/checkstyle/parser/EmptyXmlEventReader.java
rnveach: same answer as above.

patch-diff-report-tool/src/test/resources/ExpectedReportPatchOnly.html
```diff
@@ -0,0 +1,291 @@
...
+				<h2 a="Checkstyle XML difference report">Checkstyle XML difference report</h2>
+				This is symmetric difference generated from two checkstyle-result.xml reports. <br/>
+				All matching records from each XML file are deleted, then remaining records are merged into single report. <br>
+		        <a href="https://github.com/checkstyle/contribution/tree/master/patch-diff-report-tool">Utility that generated this report.</a>
+			</div>
+
+			
+
+			<div class="section">
+				<h2 a="Summary:">Summary:</h2>
```
rnveach: Fixed test to provide patch configuration, but even with that configuration difference isn't showing in report.
I will fix.

patch-diff-report-tool/README.md
```diff
@@ -15,7 +15,7 @@ You have 2 different checkstyle repos, original (base) and forked (patch), for e
 
 8) Now execute this utility with 6 command line arguments:<br/>
 
-`--baseReport` - path to the base checkstyle-result.xml (required argument);<br/>
+`--baseReport` - path to the base checkstyle-result.xml (optional argument);<br/>
```
rnveach: Done.

patch-diff-report-tool/src/main/java/com/github/checkstyle/Main.java
rnveach: Done.

patch-diff-report-tool/src/test/resources/ExpectedReportPatchOnly.html
rnveach: Change in `CliPaths.configurationPresent` needed.
Done.

rnveach:
Commit updated with some fixes.

romani:
question in issue was added

rnveach:
@romani As far as I can tell, this PR is not affected by your changes in issue.

romani:
>The utility still needs to be modified to take no base config and only the patch.

I might lost a context, but why this is required ?

rnveach:
@romani This is required for 'single' mode you described in issue.
Right now this utility can only work if 2 reports are given, base and patch, as both are required fields.
With 'single' mode, there will be no base report. Since it is a new token/check, only patch branch's report will be available.
This change makes base report an optional field now.

Issue #219, opened 2017-04-26
launch.groovy: can't handle reports with links all on 1 line


rnveach:
Identified at https://github.com/checkstyle/checkstyle/pull/4236#issuecomment-297380288 : 

Groovy script is removing all `xref` files from user's report.
When debugging the issue I see that the method, `getFilesReferencedInReport` at https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/launch.groovy#L262-L274 , is only getting one result and it is a combination of file and HTML. Each result should be just the file and not any HTML.

The issue is all the user's links are on one line and groovy is written expecting links on separate lines.

Example:
https://github.com/Kietzmann/Kietzmann.github.io/blob/master/openjdk9/site/index.html#L71

Adding a system out to `getFilesReferencedInReport` produces this one, incorrect result:
````
Removing non refernced xref files in report ...
xref/openjdk9/src/jdk.internal.le/share/classes/jdk/internal/jline/TerminalFactory.html#L156">156</a></td></tr><tr class="a"><td><img src="images/icon_warning_sml.gif" alt="" />&#160;Warning</td><td>whitespace</td><td>NoWhitespaceAfter</td><td>'::' is followed by whitespace.</td><td><a href="./xref/openjdk9/src/jdk.internal.le/share/classes/jdk/internal/jline/TerminalFactory.html#L157">157</a></td></tr></table></div><div class="section"><h3 id="openjdk9.test.sun.security.krb5.auto.principalProperty.PrincipalSystemPropTest.java">openjdk9/test/sun/security/krb5/auto/principalProperty/PrincipalSystemPropTest.java</h3><table border="0"class="bodyTable"><tr class="b"><th>Severity</th><th>Category</th><th>Rule</th><th>Message</th><th>Line</th></tr><tr class="a"><td><img src="images/icon_warning_sml.gif" alt="" />&#160;Warning</td><td>whitespace</td><td>NoWhitespaceAfter</td><td>'::' is followed by whitespace.</td><td><a href="./xref/openjdk9/test/sun/security/krb5/auto/principalProperty/PrincipalSystemPropTest.html
````

MEZk:
@romani @rnveach 
I'm on it.

romani:
as we come to idea that default HTML report from maven plugin is not enough for us, we need to start thinking on depending on XML report only and generate HTML in form that we need. Any further updates for HTML rendering in plugin will damage our tool.

MEZk:
@romani 
[...]

PR #220, opened 2017-04-26
Issue #219: Fix bug when launch.groovy cannot handle reports with links all on 1 line

MEZk:
#219 

1) Fixed regexp and problem with one line reports.
2) Replaced ArrayList with HashSet to avoid file path duplication in getFilesReferencedInReport result as index.html (checkstyle.html) can contain more than one reference to the same file.

**FYI**

The code will collect all matching substrings of the line applying the given transformer closure (```it.substring(xrefStartIdx)```):

```java
matcher.collect { it.substring(xrefStartIdx) }
```

I prepared two reports which I used for testing:
https://github.com/MEZk/mezk.github.io/tree/master/i219-launch-groovy-bug

@romani 
Codenarc panics :D
See https://travis-ci.org/checkstyle/contribution/jobs/226174048

One of the violations which references my changes:
```
   Violation: Rule=UnnecessarySubstring P=3 Line=270 Msg=[Violation in class None. The String.substring(int) method can be replaced with the subscript operator] Src=[referencedFiles.addAll(matcher.collect { it.substring(xrefStartIdx) })]
```
Personally, I don't think that usage of subscript operator of Groovy is a good idea as the code will become weird and cause confusion even to me.

### DIFF
--- a/checkstyle-tester/StarterRuleSet-AllRulesByCategory.groovy.txt
+++ b/checkstyle-tester/StarterRuleSet-AllRulesByCategory.groovy.txt
@@ -231,12 +231,16 @@ ruleset {
     ExplicitCallToPowerMethod
     ExplicitCallToRightShiftMethod
     ExplicitCallToXorMethod
-    ExplicitHashMapInstantiation
-    ExplicitHashSetInstantiation
-    ExplicitLinkedHashMapInstantiation
-    ExplicitLinkedListInstantiation
-    ExplicitStackInstantiation
-    ExplicitTreeSetInstantiation
+
+    // Suppressed as these rules make the code hard to understand.
+    // Disabling them helps to have the code more similar to Java.
+    // ExplicitHashMapInstantiation
+    // ExplicitHashSetInstantiation
+    // ExplicitLinkedHashMapInstantiation
+    // ExplicitLinkedListInstantiation
+    // ExplicitStackInstantiation
+    // ExplicitTreeSetInstantiation
+
     GStringAsMapKey
     GStringExpressionWithinString
     GetterMethodCouldBeProperty
@@ -384,7 +388,10 @@ ruleset {
     UnnecessarySelfAssignment
     UnnecessarySemicolon
     UnnecessaryStringInstantiation
-    UnnecessarySubstring
+
+    // Suppressed to avoid weired code and make the code more similar to Java.
+    // UnnecessarySubstring
+
     UnnecessaryTernaryExpression
     UnnecessaryToString
     UnnecessaryTransientModifier
--- a/checkstyle-tester/launch.groovy
+++ b/checkstyle-tester/launch.groovy
@@ -260,14 +260,14 @@ def removeNonReferencedXrefFiles(siteDir) {
 }
 
 def getFilesReferencedInReport(linesFromIndexHtml) {
-    def pattern = Pattern.compile('.*<td><a href="./xref.*\\.html#L\\d+">.*')
-    def referencedFiles = new ArrayList<String>()
+    def xrefStartIdx = 2
+    def pattern = Pattern.compile('\\./xref/[^<>]+\\.html')
+    def referencedFiles = new HashSet<String>()
     linesFromIndexHtml.each {
         line ->
             def matcher = pattern.matcher(line)
-            if (matcher.matches()) {
-                def filePath = line.substring(line.indexOf("/xref") + 1, line.lastIndexOf('#L'))
-                referencedFiles.add(filePath)
+            if (matcher.find()) {
+                referencedFiles.addAll(matcher.collect { it.substring(xrefStartIdx) })
             }
     }
     return referencedFiles
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,8 +51,8 @@ matrix:
       env:
         - DESC="codenarc validation for groovy files"
         - CMD1=" cd checkstyle-tester "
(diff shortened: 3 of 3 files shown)


checkstyle-tester/launch.groovy
```diff
@@ -260,14 +260,14 @@ def removeNonReferencedXrefFiles(siteDir) {
 }
 
 def getFilesReferencedInReport(linesFromIndexHtml) {
-    def pattern = Pattern.compile('.*<td><a href="./xref.*\\.html#L\\d+">.*')
-    def referencedFiles = new ArrayList<String>()
+    def xrefStartIdx = 2
+    def pattern = Pattern.compile('./xref/[^<>]+\\.html')
```
rnveach: Since this is an expression, should first period be `\\.`?

MEZk: Fixed.

rnveach:
Shouldn't travis be updated so it turns green then?
Maybe we should disable some of the rules we don't like. http://codenarc.sourceforge.net/codenarc-configuring-rules.html#Turning_Off_A_Rule

romani:
Travis should be green.

romani:
UnnecessarySubstring Rule - http://codenarc.sourceforge.net/codenarc-rules-unnecessary.html

I am ok to suppress.

romani:
@MEZk , please fix build and , I am ok to merge.

MEZk:
@romani 
What should I do with other violations?
https://travis-ci.org/checkstyle/contribution/jobs/226546726

Can I resolve them in separate issue, otherwise there will be too many unrelated changes.

romani:
There were no violations before your commit. You introduce new code with violations - should be addressed in the same time. Or I am missing smth ?

MEZk:
@romani 
Codenarc complains about diff.groovy, not launch.groovy

romani:
@MEZk , take a look at .travis.yml .... "diff.log && cat diff.log && grep '(p1=0; p2=22; p3=135)' diff.log" 
in your buid "(p1=0; p2=22; p3=133)", so your code fixed two issues, so you need to update travis to pass a build.

romani:
@MEZk , @rnveach , "def should not be use" is good if we plan to keep Groovy in close to java syntax to ease code copy-paste.
Please share your ideas on this. My vote is it keep this rule.

MEZk:
@romani 
It will mean that we cannot use Groovy syntax at all as we should keep in mind that someday we may need to copy-paste the script to Java.

MEZk:
@romani 
> so your code fixed two issues,

I fixed nothing in diff.groovy.

romani:
> I suggest disabling ExplicitHashSetInstantiation Rule

I am agree to disable all such rules to be a bit verbose to keep code similar to java.
Please do.

MEZk:
@romani 
> I am agree to disable all such rules to be a bit verbose to keep code similar to java.
Please do.

Done. I suggest discussing other tules in the scope of https://github.com/checkstyle/contribution/issues/182. The results of the discussion will be applied to  StarterRuleSet-AllRulesByCategory.groovy.txt.

checkstyle-tester/StarterRuleSet-AllRulesByCategory.groovy.txt
```diff
@@ -231,12 +231,6 @@ ruleset {
     ExplicitCallToPowerMethod
     ExplicitCallToRightShiftMethod
     ExplicitCallToXorMethod
-    ExplicitHashMapInstantiation
-    ExplicitHashSetInstantiation
-    ExplicitLinkedHashMapInstantiation
-    ExplicitLinkedListInstantiation
-    ExplicitStackInstantiation
-    ExplicitTreeSetInstantiation
```
romani: please use commenting instead of removal and define reason of suppression

MEZk: Done

checkstyle-tester/StarterRuleSet-AllRulesByCategory.groovy.txt
```diff
@@ -384,7 +388,10 @@ ruleset {
     UnnecessarySelfAssignment
     UnnecessarySemicolon
     UnnecessaryStringInstantiation
+
+    // Suppressed to avoid weired code and make the code more similar to Java.
     UnnecessarySubstring
+
```
romani: I do not see suppression

MEZk: Fixed

checkstyle-tester/StarterRuleSet-AllRulesByCategory.groovy.txt
```diff
@@ -231,12 +231,16 @@ ruleset {
...
     ExplicitCallToXorMethod
-    ExplicitHashMapInstantiation
-    ExplicitHashSetInstantiation
-    ExplicitLinkedHashMapInstantiation
-    ExplicitLinkedListInstantiation
-    ExplicitStackInstantiation
-    ExplicitTreeSetInstantiation
+
+    // Suppressed as these rules make the code hard to understand.
+    // Disabling them haelps to have the code more similar to Java.
```
rnveach: `haelps` -> `helps`

MEZk: Fixed

PR #228, opened 2017-05-10
minor: removed empty files with no violations from patch only report

rnveach:
Fixed issue where empty files were being displayed in patch only report. This polluted the report with too many headers in a real scenario.
See http://rveach.no-ip.org/checkstyle/regression/reports/30/checkstyle/index.html

Fix ignores files with no violations. Test regression shows it only affects patch only report with less numbers and headers. It doesn't affect the overall file number count.

Will rename commit to reference this pull number.


rnveach:
Commit was renamed.

Issue #240, opened 2017-06-02
patch-diff-report-tool: support non-java file xrefs


rnveach:
This issue will become more common with #239 .

We currently only use JXR to create the xrefs and it requires the file to be an almost valid Java file.
If the file isn't valid, than we get no XREF as seen with test report https://github.com/checkstyle/contribution/blob/0c13b1642332adfb98521d1b6dee19363b5a82d5/patch-diff-report-tool/src/test/resources/ExpectedReportNonCompilable.html#L84 .

Issue happens at https://github.com/checkstyle/contribution/blob/0c13b1642332adfb98521d1b6dee19363b5a82d5/patch-diff-report-tool/src/main/java/com/github/checkstyle/site/XrefGenerator.java#L126 when JXR throws an exception.

We need to create an alternative to see xref files of non-Java files and non-compilable Java files.

rnveach:
Fix was merged

PR #241, opened 2017-06-02
Issue #240: added support for xrefs that fail JXR

rnveach:
Issue #240

Copied `JavaCodeTransform`, trimmed it down to remove Java related stuff, and made it pass checkstyle.
Added a test to show what the output of `TextTransform` will look like.

I question if we should just remove JXR completely, but it does bold keywords and such which is slightly nice.


romani:
> I question if we should just remove JXR completely, but it does bold keywords and such which is slightly nice.

Lets put here some report with keyword highlihts and withtout and do a vote.

PR #242, opened 2017-06-02
config: changed patch-diff-report-tool to verify test sources style

rnveach:
Changing config to run checkstyle on sources directory.
I expect 1 violation.


rnveach:
Violation was given and fixed.

Issue #243, opened 2017-06-04
releasenotes-builder: Extra "\n" in release notes generations


romani:
https://travis-ci.org/checkstyle/checkstyle/jobs/237007609 look at raw log

it force me to extra updates to notes to copy them posts or xdoc or .....

Vladlis:
@romani 
your link gives me
```xml
<Error>
<Code>AccessDenied</Code>
<Message>Request has expired</Message>
<X-Amz-Expires>30</X-Amz-Expires>
<Expires>2017-06-03T20:50:39Z</Expires>
<ServerTime>2017-06-04T18:12:42Z</ServerTime>
<RequestId>ECE0674480B81BC0</RequestId>
<HostId>XX/BIdZJh0Rq5Yg77XYoqvoAtiDqVJeBZaHtThEroSFXOCfb8erzLaSrwufEhEFdrO2XM8gyKW4=</HostId>
</Error>
```

romani:
The same it gives to me now, I updated link in description to Travis logs, looks like raw log link is not persistent

MEZk:
@romani 
Thymeleaf is sensitive to lines formatting. [This](https://github.com/checkstyle/contribution/blob/master/releasenotes-builder/src/main/resources/com/github/checkstyle/templates/xdoc_thymeleaf.template#L7) line should have the same line number as the previous one. We will continue getting similar issues till we move to freemarker.

romani:
Please convert us to freemarker, this task is not for Thymeleaf

MEZk:
Switch to the freemarker will be done in the scope of https://github.com/checkstyle/contribution/issues/212

romani:
@MEZk , I do think that we should use only freemarker for this file generation.
It was my design mistake to try to use thymeleaf for xdoc.
thymeleaf is good for html files, that helps to keep template looks well in browser, and result page rendering is well, nobody care about clear html code.
Xdoc - is XML file with script formatting of code, so freemarker is better tool by design for this task.

PR #244, opened 2017-06-08
Issue #243: Fix thymeleaf template to avoid incorrect placement of '\n'

MEZk:
issue #243 

Fixed thymeleaft template to have it formatted correct in case freemarker has problems. I'll switch the tool to freemarker in the next PR.

Template looks wired, but works.

### DIFF
--- a/releasenotes-builder/pom.xml
+++ b/releasenotes-builder/pom.xml
@@ -25,6 +25,7 @@
         <tools.jar.version>${java.version}.0</tools.jar.version>
         <tools.jar.path>${java.home}/../lib/tools.jar</tools.jar.path>
         <checkstyle.version>7.8.1</checkstyle.version>
+        <junit.version>4.12</junit.version>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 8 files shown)


rnveach:
@MEZk Since this is a Java project and we can't see example of results with this change, maybe we should consider sometime adding a test that will print out a report for us like we do for patch-diff-report-tool and it's tests?

MEZk:
@rnveach 
I opened the issue https://github.com/checkstyle/contribution/issues/245.

MEZk:
No need in the new issue, I'll add UTs in the scope of this one.

MEZk:
@rnveach 
Added UTs.

rnveach:
@MEZk thanks.

romani:
@MEZk , TC is not happy.

MEZk:
@romani 
Fixed.

MEZk:
@romani 
Please, check and merge this PR to let me reuse the test class and switch us to freemarker. Fix is ready.

romani:
I am ok with change if it works.

Issue #237, opened 2017-05-29
Regression Testing tool for javadoc parsing.


ps-sp:
UTs for javadocs are insufficient. Also they currently have a poor implementation style ([4381](https://github.com/checkstyle/checkstyle/issues/4381)).

Thus, regression testing is a must for most of the grammar updates. The parse trees in the master and the patch should be compared and the differences should be reported.

Vladlis:
The initial thought was to make a simple script to do comparison of CLI output, but results would have been not representative and hard to analyze.

[Proposal](https://github.com/checkstyle/checkstyle/pull/4378#issuecomment-304561217) from @rnveach is to create a utility check that would only print AST as a violation so that we could use the existing checkstyle-tester for reports. I think this makes more sense even though
> the XML files (temporary report files) may be larger

For me it is not clear how are we going to use such a check:
- locally add it to the project every time (unlikely)
- add it as a utility to the main project?
- add like a third party check?

@romani @rnveach @sabaka Please join the conversation.

rnveach:
> simple script to do comparison of CLI output

This was my second plan, if the first didn't work out.
One issue is the main goal was to compare the Javadoc structure. To use existing projects for regression, we need to first identify the Javadocs in the files. The CLI currently can't do this unless we use the `-J` option and cut out the non-Javadoc syntax, which would require a completely new process.
Another issue is regression needs to be able to show the original Java file in the web so we know what the differences are we are looking at. CLI doesn't do this, so we would need to apply JXR from normal regression to this new process somehow.
We would also need a utility to find similarities and differences, even across multiple lines, as new nodes would cause existing lines to be moved down. patch-diff-report-util can't handle this type of comparison.

There are some issues with the way I implemented the check, which may also cause some issues for regression.
Line/column numbers won't match up to the actual line/column numbers in the file. I made a hack just to make sure the line numbers are unique, as we sort the results based on those and incorrect sorting will make the printed tree hard to review.
Also because of the above, new/removed lines in the differences will affect all lines after it because line numbers is based on AST output line number. It won't see the same text on different lines as the same violation. This would still be an issue with the CLI way too and I said with it.
Doing all the files in the project will create a lot of violations, and may make the final reports large depending on the number of changes. I saw a huge slowdown in the openjdks projects, when testing this out, so I recommend avoiding them.

> add like a third party check

This seems like the best option to me. We would have our own set of checks for testing regression.

romani:
[...]

PR #247, opened 2017-06-10
Issue #237: added antlr regression script

rnveach:
Issue #237

Very similar to the layout of my other script.

Runs checkstyle's `-J` option on all java files in a project. It does 2 files in parallel, base and patch.
This produces java and javadoc trees so we can compare all ANTLR in one script.

I left a `echo "$f"` display in the code to verify where the script is in it's progress because of how long it takes to do an entire project.

Checkstyle: 34 minutes
Sevntu: 3 minutes
guava: 26 minutes

### DIFF
--- a/checkstyle-tester/launch_diff_antlr.sh
+++ b/checkstyle-tester/launch_diff_antlr.sh
@@ -1,0 +1,319 @@
+#!/bin/bash
+
+source launch_diff_variables.sh
+
+EXTPROJECTS=()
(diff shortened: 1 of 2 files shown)


romani:
looks like @rnveach , become experienced in shell scripting :) . One day he will switch to development completely on Linux.

rnveach:
> One day he will switch to development completely on Linux.

That gave me a good laugh. :)

Issue #265, opened 2017-09-14
Checkstyle-tester: incorrect work launch.groovy for Windows


sergileon:
I generated report uses diff.groovy, but it uses [https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/diff.groovy#L190](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/diff.groovy#L190). [Launch.groovy](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/launch.groovy#L243) uses a method `getFilesReferencedInReport` that returns HashSet with incorrect paths for Windows. Windows uses "\\" for path, but method return "/".
Screenshot in debug:
![1232313123](https://user-images.githubusercontent.com/19771549/30421661-1f729c04-9946-11e7-9066-7dd4150df010.png)

romani:
@rnveach , please take a look. 
Strange that it worked before, there were bunch of contributors with Windows who generated correct reports.

sergileon:
@romani , now I description the principle work generated report. If I'm wrong, correct me, please. In begin call method [generateCheckstyleReport](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/diff.groovy#L28). Inside, call [launch.groovy](https://github.com/sergileon/contribution/blob/master/checkstyle-tester/diff.groovy#L190). After, `launch.groovy` generated report and clear folder [src/main](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/launch.groovy#L87)
Generated report contain file `checkstyle-result.xml`.
Example  `checkstyle-result.xml` on Windows:
```
<?xml version="1.0" encoding="UTF-8"?>
<checkstyle version="8.3-SNAPSHOT">
<file name="D:\project\check\contribution\contribution\checkstyle-tester\src\main\java\guava\guava\src\com\google\common\annotations\Beta.java">
</file>
<file name="D:\project\check\contribution\contribution\checkstyle-tester\src\main\java\guava\guava\src\com\google\common\annotations\GwtCompatible.java">
</file>
<file name="D:\project\check\contribution\contribution\checkstyle-tester\src\main\java\guava\guava\src\com\google\common\annotations\GwtIncompatible.java">
</file>
<file name="D:\project\check\contribution\contribution\checkstyle-tester\src\main\java\guava\guava\src\com\google\common\annotations\package-info.java">
</file>
[...]

PR #266, opened 2017-09-15
Issue #265:fixed incorrect path for Windows

sergileon:
fixed [issue](https://github.com/checkstyle/contribution/issues/265)


romani:
@sergileon , codenarc is not happy. Amount of violations should only decrease or stay the same.

sergileon:
@romani, codenarc is ok.

romani:
Greb was simple solution, when we tried to enforce compliance with codenarc. Everyone are welcome to make it better

Issue #273, opened 2017-12-25
Launch/Diff Groovy should remove use of maven-checkstyle-plugin


rnveach:
See https://issues.apache.org/jira/browse/MCHECKSTYLE-346 .
We need to deprecate all usage of `maven-checkstyle-plugin` so we can start implementing backward breaking changes.

The groovy scripts are our heaviest scripts as we use them for all types of regression, so they should be the first ones converted.

Instead of installing checkstyle and then running `mvn site`, we should package the project into the `all` jar and run the CLI checkstyle program and produce the XML violation file. This file and the sources will be what is fed into `patch-diff-report-tool` to produce the final report.

We should not need to change `patch-diff-report-tool`. Using the CLI allows us to run Checkstyle on the repo directory directly instead of copying the files to another location temporarily. This will also allow us to run regression on non-Java files which is what `maven-jxr-plugin` forced on us.

romani:
I am ok create issue on Checkstyle to generate HTML report, I think we already have such issue, and raised this in discussions several times.
Or users/me can use xsl to convert xml to html (very old fashioned approach).

Html generation is better to be done html template libraries but  it is new dependency .... so it is ok to generate html by old-fashioned way (StringBuilder.append)

rnveach:
> I am ok create issue on Checkstyle to generate HTML repor

For regression, we don't need this. `patch-diff-report-tool` creates the final HTML report for us. We don't use any output of `maven-jxr-plugin` for anything.

romani:
>we don't use the output of maven-jxr-plugin for anything. 

this is result of maven-jxr-plugin - 
https://djydewang.github.io/diffReport_Issue4981/findbugs-with-excldues/xref/home/bbg/project/contribution/checkstyle-tester/repositories/findbugs-with-excldues/eclipsePlugin/src/edu/umd/cs/findbugs/plugin/eclipse/quickfix/util/SourceLineVisitor.java.html#L26 , without generation of such pages, html report is usless for us , to my mind.
We should be able to run it on sources (with fake pom.xml) to generate folder of html pages.

rnveach:
> this is result of maven-jxr-plugin 

`patch-diff-report-tool` generates this via `JavaCodeTransform` and `TextTransform`. This is why we have these code transforms in the tool.
https://github.com/checkstyle/contribution/blob/master/patch-diff-report-tool/src/main/java/com/github/checkstyle/site/XrefGenerator.java#L136-L160
[...]

PR #274, opened 2017-12-26
launch_diff: upgraded util to not rely on maven plugin

rnveach:
To show how https://github.com/checkstyle/contribution/issues/273 should be done, I updated my own script to not rely on maven anymore.
It uses the pure Checkstyle CLI with no updates to patch-diff-report-tool needed and not extra dependencies.

* All `install` commands where changed to `package` now.
* `github` support was removed, should just rely on `git`.
* added extra code to prevent switching to `master` as if it was a different branch.
* added branch names and commit ids to final report like groovy scripts.

Here is an example run of the differences for the same PR and config:

Maven: http://rveach.no-ip.org/checkstyle/regression/reports/149/
CLI: http://rveach.no-ip.org/checkstyle/regression/reports/150/

Only real difference is counts for number of files being processed overall. Maven was restricting us to Java only file types while CLI pulls in all files.
No changes in difference reporting. No changes to internal web page acting as interface for user to program.
Slight speed increase as we don't need to copy large quantity of files around (twice) or try to minimize those files. Less output to console too (263kb versus 103kb).

An external change will be needed to switch to using CLI full time. Excludes uses a wildcard match while CLI excludes must be a pattern matching. So all excludes will need to be updated in scope of https://github.com/checkstyle/contribution/issues/273 .

### DIFF
--- a/checkstyle-tester/launch_diff.sh
+++ b/checkstyle-tester/launch_diff.sh
@@ -3,9 +3,9 @@
 source launch_diff_variables.sh
 
 EXTPROJECTS=()
-INSTALL_MASTER=true
+PACKAGE_MASTER=true
 RUN_MASTER=true
-INSTALL_PULL=true
+PACKAGE_PULL=true
 RUN_PULL=true
 RUN_REPORTS=true
 USE_CUSTOM_CONFIG=false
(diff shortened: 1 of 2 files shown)


rnveach:
I am unsure what travis failure is about.

romani:
Mi be some network issue, restarted.

romani:
Ones CIs resolved ok to merge.

rnveach:
@romani TC failure is from releasenotes-builder .
Travis is passing now.

rnveach:
@romani TC failures look false.

> Unused import (64)

None of these are true.

> Declaration has problems in Javadoc references (9)

All these are exceptions and can be found by the compiler.

> Redundant throws clause (Errors) (4)

These exceptions are thrown by the utilities.

rnveach:
TC has passed after a re-run.

PR #297, opened 2018-04-15
minor: migrate exclusion code from checkstyle

jsoref:
for checkstyle/checkstyle#5719


romani:
it is not good to bind repositories that much, but I ma ok , to make it work first and then optimize.

PR #301, opened 2018-06-07
sevntu_launch_diff: added sevntu difference utility

rnveach:
Like checkstyle one, it doesn't use maven.

To run sevntu easily, I have it copy itself over to Checkstyle repo and package it that way.
I recently had to update this check to only use release version of checkstyle as master contains backwards breaking changes that conflict with sevntu's current checkstyle version.

Otherwise there is not much difference between this util and the checkstyle one.


romani:
please share example of use-cases and examples of commands usage

Issue #302, opened 2018-06-09
launch.groovy: option to override sevntu version


rnveach:
Identified at https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/644#issuecomment-395990184 , we need an option to override sevntu version to configure no exception testing for it.

rnveach:
Fix was merged

PR #303, opened 2018-06-09
Issue #302: added sevntu version option to launch groovy

rnveach:
Issue #302


romani:
merged to unblock other work 

BUT please do not increase amount of violations from codenarc, it always should go done, and eventually come to "0". `p2=21; p3=1` is our technical dept.

Issue #309, opened 2018-06-23
releasenotes-builder: allow other repositories


rnveach:
Talked about at https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/638#issuecomment-357542972 ,

We should remove any `checkstyle` hardcoding and allow other repositories, like sevntu, to use our tool to build their release notes.

First change I see we needing is to remove
https://github.com/checkstyle/contribution/blob/d9836c08428486e758d6950bf972256e60859171/releasenotes-builder/src/main/java/com/github/checkstyle/Main.java#L45-L46
And allow users to specify remote github location through CLI.

romani:
@rnveach , releasenotes generation process is very different from project to project.
What if we will move required parts to util methods(or smth another ....) to be reused in new builder ?

rnveach:
Right now my main aim was HTML release notes with links to issues and type of change.

> releasenotes generation process is very different from project to project.

Can you give more specific scenarios? You can use checkstyle and sevntu.

romani:
I am ok with all changes.

rnveach:
Talked about at https://github.com/checkstyle/contribution/pull/310#issuecomment-399729915 ,

we need an extra options to allow user to supply custom xdoc template and custom xdoc location.

rnveach:
sevntu used release notes just fine, so I am closing this issue as I don't see anything else to add.

PR #315, opened 2018-07-01
Issue #309: allow remote repository path variable in freemarker

rnveach:
Issue #309

It is a required field, so we should allow it to be used in templates.

### DIFF
--- a/releasenotes-builder/src/main/resources/com/github/checkstyle/templates/xdoc_freemarker.template
+++ b/releasenotes-builder/src/main/resources/com/github/checkstyle/templates/xdoc_freemarker.template
@@ -5,7 +5,7 @@
         <ul>
         <#list breakingMessages as message>
           <li>
-            ${message.title}. Author: ${message.author}<#if message.issueNo != -1> <a href="https://github.com/checkstyle/checkstyle/issues/${message.issueNo}">#${message.issueNo}</a></#if>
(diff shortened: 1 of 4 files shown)


releasenotes-builder/src/main/resources/com/github/checkstyle/templates/xdoc_freemarker.template
```diff
@@ -5,7 +5,7 @@
         <ul>
         <#list breakingMessages as message>
           <li>
-            ${message.title}. Author: ${message.author}<#if message.issueNo != -1> <a href="https://github.com/checkstyle/checkstyle/issues/${message.issueNo}">#${message.issueNo}</a></#if>
+            ${message.title}. Author: ${message.author}<#if message.issueNo != -1> <a href="https://github.com/${remoteRepoPath}/issues/${message.issueNo}">#${message.issueNo}</a></#if>
```
romani: in next PR, please add "\n" here make line about 100 symbols and be ease to review
it will also resolve our problem with generation of long lines in result XML.

rnveach: yes, this is a separate issue, so I am waiting for that to do it.
I did this for sevntu since it is a new template,

PR #321, opened 2018-07-02
minor: fixed wrong template resource locations

rnveach:
Fix for https://travis-ci.org/checkstyle/checkstyle/jobs/398967176#L741


romani:
Good question: Why our tests did not cought such problem ?

rnveach:
Our tests don't call Main, so they aren't fully testing all changes. This utility needs more and better tests.

Issue #327, opened 2018-07-06
releasenotes-builder: add better in-depth tests


rnveach:
We need few tests for parsing and creation of release notes content that is based on Github issue titles that are referenced in commits.

romani:
@rnveach , do you want to add more tests ?

rnveach:
Yes, I want to add some tests for the guy title parsing.

romani:
we can try to use mockserver https://www.mock-server.com/mock_server/running_mock_server.html
to simulate responses from Github api.
1-2 test methods is enough.

I do not think we should deal with real/mock of git repo. we can just simulate that jgit library return us some results that we need.
1-2 test methods is enough.

romani:
several tests were done at https://github.com/checkstyle/contribution/pull/705

@stoyanK7 , to close this issue, please convert existing test for content validation to use this approach of testing.
It will be enough for this issue to be closed.

stoyanK7:
On it.

rnveach:
Fix was merged

PR #328, opened 2018-07-06
Issue #327: added template tests that hit more main code

rnveach:
Issue #327

This PR is aimed more at template processing. I will make another one for commit parsing.

Most code was moved from `Main` to `MainProcess` so we could hit the testable code easier.
Fixed an invalid label name.
Added more tests.
Fixed bug where we swapper remote repo path and release number.


rnveach:
@romani It looks like IDEA changed something? I tried re-running it and it failed the same way.

romani:
from Teamcity logs:

> [Inspection output] This IDEA distribution is only licensed to be used with JetBrains TeamCity.

smth is broken on their side.

PR #345, opened 2019-01-14
minor: bump year to 2019 in headers

romani:
```
$ git diff | grep -Eh "^\+"  | grep -v "+++ b" | sort | uniq
+// Copyright (C) 2001-2019 the original author or authors.
```


rnveach:
> Execution check of goal org.apache.maven.plugins:maven-checkstyle-plugin:2.17:check failed: String index out of range: -1

There is a violation in the project from a check that has an ID specified.

PR #346, opened 2019-03-08
config: remove AnnotationLocation token PARAMETER_DEF as it will be removed from Checkstyle for #6462

Vampire:
Related to issue https://github.com/checkstyle/checkstyle/issues/6462 and PR https://github.com/checkstyle/checkstyle/pull/6530

### DIFF
--- a/checkstyle-tester/checks-nonjavadoc-error.xml
+++ b/checkstyle-tester/checks-nonjavadoc-error.xml
@@ -133,12 +133,29 @@
 
     <!-- Annotations -->
     <module name="AnnotationLocation">
+      <property name="tokens" value="CLASS_DEF"/>
+      <property name="tokens" value="INTERFACE_DEF"/>
+      <property name="tokens" value="PACKAGE_DEF"/>
+      <property name="tokens" value="ENUM_CONSTANT_DEF"/>
+      <property name="tokens" value="ENUM_DEF"/>
+      <property name="tokens" value="METHOD_DEF"/>
+      <property name="tokens" value="CTOR_DEF"/>
+      <property name="tokens" value="VARIABLE_DEF"/>
       <property name="tokens" value="ANNOTATION_DEF"/>
       <property name="tokens" value="ANNOTATION_FIELD_DEF"/>
       <property name="allowSamelineSingleParameterlessAnnotation" value="false"/>
     </module>
     <module name="AnnotationLocation">
-      <property name="tokens" value="PARAMETER_DEF"/>
+      <property name="tokens" value="CLASS_DEF"/>
+      <property name="tokens" value="INTERFACE_DEF"/>
+      <property name="tokens" value="PACKAGE_DEF"/>
+      <property name="tokens" value="ENUM_CONSTANT_DEF"/>
+      <property name="tokens" value="ENUM_DEF"/>
+      <property name="tokens" value="METHOD_DEF"/>
+      <property name="tokens" value="CTOR_DEF"/>
+      <property name="tokens" value="VARIABLE_DEF"/>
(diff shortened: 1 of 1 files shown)


checkstyle-tester/checks-nonjavadoc-error.xml
```diff
@@ -139,10 +139,6 @@
       <property name="tokens" value="ANNOTATION_FIELD_DEF"/>
       <property name="allowSamelineSingleParameterlessAnnotation" value="false"/>
     </module>
-    <module name="AnnotationLocation">
-      <property name="tokens" value="PARAMETER_DEF"/>
-      <property name="allowSamelineMultipleAnnotations" value="true"/>
```
rnveach: http://checkstyle.sourceforge.net/config_annotation.html#AnnotationLocation_Properties
Default value is `false`. I think we should leave the module in there and just remove the token so this property is still tested.

Vampire: With which tokens?
Default?
Do you think it will then work properly?

rnveach: Yes, just remove tokens so it runs with default and keep allowSamelineMultipleAnnotations to true.
It shouldn't be an issue, older versions will still work as it is the loading of the check in `TreeWalker` that defines what tokens will be used for default.

Vampire: I just had concerns that then checks are run that were not run before and could cause breakage (as all default tokens are tested instead of just the parameter def), but if you say so, I'm fine, fixed.

rnveach: If there is any breakage, then it will show us another problem in the check.
Checks should not throw an exception with any optional tokens added/removed. If a token is required for the check to work correctly, then it should be listed as required.

Vampire: I more thought about new violations found that are not expected.
But maybe I misinterpreted the intention of these regression checks, I didn't look closer into them.

rnveach: This specific check is for non-exception testing. 
Severity is set to ignore, so no violations are printed except exceptions.
https://github.com/checkstyle/contribution/blob/46221abc3b446b0826d363853cccb2028d16cadc/checkstyle-tester/checks-nonjavadoc-error.xml#L17

When we use project's own config in regression, it is violation testing. So we expect no new violations as long as they are strict in using checkstyle before adding new commits to their own repo.

Vampire: Ah, ok, that makes sense, thanks.

Vampire:
@romani ping :-)

checkstyle-tester/checks-nonjavadoc-error.xml
```diff
@@ -138,7 +138,6 @@
       <property name="allowSamelineSingleParameterlessAnnotation" value="false"/>
     </module>
     <module name="AnnotationLocation">
-      <property name="tokens" value="PARAMETER_DEF"/>
```
romani: I am ok to remove such tokentype, ....
but lets use here all tokens - https://checkstyle.org/config_annotation.html#AnnotationLocation (except for PARAMETER_DEF)

whole reason of this config it use all possible tokens/modes and detect unexpected exceptions of massive code.

@rnveach , does it make sense ?

rnveach: @romani I agree.
Since we are updating to use all tokens, we should do the same with the other `AnnotationLocation` config right above it.

Vampire: done

PR #358, opened 2019-03-17
Issue #355: add some documentation for local type

### LINKED ISSUE #355: local repos in checkstyle-tester
The documentation of checkstyle-tester suggests it once supported local paths for repos.
While this is still possible when it is a local Git or Hg repository (example gets fixed in #354), it would be nice to support a "local" type to support any VCS including none, local repositories with modified working trees and it would also save the space in `repositories`.

Actually what I suppose ist just this which I could PR if you like it:

```diff
-                cloneRepository(repoName, repoType, repoUrl, commitId, reposDir)
                 deleteDir(srcDir)
-                copyDir(getOsSpecificPath("$reposDir", "$repoName"), getOsSpecificPath("$srcDir", "$repoName"))
+                if (repoType == 'local') {
+                    copyDir(repoUrl, getOsSpecificPath("$srcDir", "$repoName"))
+                } else {
+                    cloneRepository(repoName, repoType, repoUrl, commitId, reposDir)
+                    copyDir(getOsSpecificPath("$reposDir", "$repoName"), getOsSpecificPath("$srcDir", "$repoName"))
+                }
```

Vampire:
Closes #355

### DIFF
--- a/checkstyle-tester/README.md
+++ b/checkstyle-tester/README.md
@@ -85,11 +85,16 @@ ATTENTION:
 
 Administrators recommend you modify `projects-to-test-on.properties` and test as many projects as possible. Each project has its own unique style and it is common to find regression in 1 and not the others.
 
-You can specify at projects-to-test-on.properties, a path to a local file system's git or hg repository to run regressions on any projects you have personally saved. See the following example:
+You can also specify projects that are already available on your local file system in `projects-to-test-on.properties`.
+For this you can either use `git` or `hg` type which will clone the local repository into the workspace and use the specified branch.
+Alternatively you can use `local` type, where the specified branch is ignored and the current state on the disk is used as is.
+The latter does not depend on any specific version control being used.
+See the following examples:
 
 ```
+my_custom_project|local|/home/username/java/my_custom_project||
 my_custom_checkstyle|git|/home/username/java/git-repos/checkstyle/checkstyle|master|**/ignoreFolderOrFiles/**/*
-my_custom_repo|hg|/home/username/java/hg-repos/myRepo|default||
+my_custom_repo|hg|/home/username/java/hg-repos/myRepo|default|
 ```
 
 ## Checkstyle pitest regression:


checkstyle-tester/README.md
```diff
@@ -85,11 +85,16 @@ ATTENTION:
...
 
-You can specify at projects-to-test-on.properties, a path to a local file system's git or hg repository to run regressions on any projects you have personally saved. See the following example:
+You can also specify projects that are already available on your local file system in `projects-to-test-on.properties`.
+For this you can either use `git` or `hg` type which will clone the local repository into the workspace and use the specified branch.
+Alternatively you can use `local` type, where the specified branch is ignored and the current state on the disk is used as is.
+The latter does not depend on any VCS being available.
+See the following examples:
 
 ```
+my_custom_project|local|/home/username/java/git-repos/my_custom_project||
```
rnveach: I wouldn't have this example be in `git-repos` to avoid confusing this with the `git` option.

checkstyle-tester/README.md
```diff
@@ -85,11 +85,16 @@ ATTENTION:
 
 Administrators recommend you modify `projects-to-test-on.properties` and test as many projects as possible. Each project has its own unique style and it is common to find regression in 1 and not the others.
 
-You can specify at projects-to-test-on.properties, a path to a local file system's git or hg repository to run regressions on any projects you have personally saved. See the following example:
+You can also specify projects that are already available on your local file system in `projects-to-test-on.properties`.
+For this you can either use `git` or `hg` type which will clone the local repository into the workspace and use the specified branch.
+Alternatively you can use `local` type, where the specified branch is ignored and the current state on the disk is used as is.
+The latter does not depend on any VCS being available.
```
rnveach: instead of `VCS`, I would the more simple term like `server`.

checkstyle-tester/README.md
Vampire: Well, it is pretty valid to reference a local Git directory too with the `local` option as it then just takes the current working directory as is, including any local modifications.
But if you think it is too confusing, ok, I removed the `git-repos/` from the example.

checkstyle-tester/README.md
Vampire: Well, with `git` or `hg` with a local repository there also needs no `server` to be available as it is a local clone that is done.
Maybe better `The latter does not depend on any specific version control being used.`?

rnveach: Yes that sounds fine to me.

PR #360, opened 2019-03-17
checkstyle-tester: do not fail if analysed project does not contain Java files because no report is present then

Vampire:
When I batch-added all my 170 locally available projects, there were some without Java files.
For those no report was generated and then the removeNonReferencedXrefFiles step failed with FileNotFoundException.
With this change, this is simply ignored.

### DIFF
--- a/checkstyle-tester/launch.groovy
+++ b/checkstyle-tester/launch.groovy
@@ -259,23 +259,26 @@ def removeNonReferencedXrefFiles(siteDir) {
     def linesFromIndexHtml = Files.readAllLines(Paths.get("$siteDir/index.html"))
     def filesReferencedInReport = getFilesReferencedInReport(linesFromIndexHtml)
 
-    Paths.get(getOsSpecificPath("$siteDir", "xref")).toFile().eachFileRecurse {
-        fileObj ->
-            def path = fileObj.path
-            path = path.substring(path.indexOf("xref"))
-            if (isWindows()) {
-                path = path.replace("\\", "/")
-            }
-            def fileName = fileObj.name
-            if (fileObj.isFile()
+    def xrefPath = Paths.get(getOsSpecificPath("$siteDir", "xref"))
+    if (Files.isDirectory(xrefPath)) {
+        xrefPath.toFile().eachFileRecurse {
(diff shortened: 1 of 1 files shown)


checkstyle-tester/launch.groovy
```diff
@@ -259,23 +259,26 @@ def removeNonReferencedXrefFiles(siteDir) {
...
-        fileObj ->
-            def path = fileObj.path
-            path = path.substring(path.indexOf("xref"))
-            if (isWindows()) {
-                path = path.replace("\\", "/")
-            }
-            def fileName = fileObj.name
-            if (fileObj.isFile()
+    def xrefPath = Paths.get(getOsSpecificPath("$siteDir", "xref"))
+    if (Files.isDirectory(xrefPath)) {
```
rnveach: Its not clear from diff but these are the only 2 new lines.

rnveach: @Vampire How does diff.groovy handle xref not existing for a project?
Does it still generate the final report correctly?

Vampire: I just added a check that the `xref` directory exists.
Because if there are no investigated files (e. g. running `AnnotationLocation` check and there are no Java files present), this directory does not exists.
If this happens, the `launch.groovy` fails with a `FileNotFoundException` and aborts.
With the check that it exists, it goes on and as far as I could see it worked fine.
I've just run a regression check on 191 projects last night (most of the ones in `projects-to-test-on.properties` plus all the ones I have on my local disk) and with this change it run through without problem.
I'm not sure what you mean with `xref not existing for a project`.
This method I modified looks through the xrefs and deletes unreferenced ones, but as the directory did not exist, the `eachFileRecurse` method failed with `FileNotFoundException`.
But if there is no `xref` directory, there is also nothing unreferenced to delete, so I think it is ok to just skip the step in this case.

rnveach: i just meant, before you got an exception and now you don't. Does diff.groovy continue on also and has no problems with the final report even though no files were processed.

I confirmed diff.groovy finishes and report looks fine.

PR #368, opened 2019-04-08
Issue #362: show separately how many violations were added and removed

### LINKED ISSUE #362: difference counts in regression diff report are wrong
The difference values of the diff report are wrong:
grafik
When fixing this, it could also be evaluated to also show the difference with `+` and `-` prefix.

Vampire:
Fixes #362

### DIFF
--- a/checkstyle-tester/diff.groovy
+++ b/checkstyle-tester/diff.groovy
@@ -1,6 +1,7 @@
 import static java.lang.System.err
 
 import java.nio.file.Paths
+import java.util.regex.Pattern
 
 static void main(String[] args) {
     def cliOptions = getCliOptions(args)
@@ -321,9 +322,10 @@ def getProjectsStatistic(diffDir) {
                 def indexHtmlFile = new File(fileObjf.absolutePath + '/index.html')
                 indexHtmlFile.eachLine {
                     line ->
-                        if (line.matches(".*totalDiff\">[0-9]+.*")) {
-                            def totalDiff = Integer.valueOf(line.substring(line.indexOf('>') + 1,
-                                    line.lastIndexOf('<')))
+                        def linePattern = Pattern.compile("totalDiff\">(?<totalDiff>[0-9]++)")
+                        def lineMatcher = linePattern.matcher(line)
+                        if (lineMatcher.find()) {
+                            def totalDiff = Integer.valueOf(lineMatcher.group('totalDiff'))
                             projectsStatistic.put(projectName, totalDiff)
                         }
                 }
--- a/patch-diff-report-tool/config/import-control.xml
+++ b/patch-diff-report-tool/config/import-control.xml
@@ -1,13 +1,14 @@
 <?xml version="1.0"?>
 <!DOCTYPE import-control PUBLIC
-        "-//Puppy Crawl//DTD Import Control 1.1//EN"
-        "http://www.puppycrawl.com/dtds/import_control_1_1.dtd">
+        "-//Puppy Crawl//DTD Import Control 1.4//EN"
+        "http://www.puppycrawl.com/dtds/import_control_1_4.dtd">
 
 <import-control pkg="com.github.checkstyle">
     <allow pkg="java.io" />
     <allow pkg="java.nio" />
     <allow pkg="java.util" />
     <allow pkg="java.text" />
+    <allow pkg="java.math" />
     <allow pkg="com.github.checkstyle.data" />
     <allow pkg="com.github.checkstyle" />
     <allow pkg="org.apache.commons.cli" />
--- a/patch-diff-report-tool/src/main/java/com/github/checkstyle/data/DiffReport.java
+++ b/patch-diff-report-tool/src/main/java/com/github/checkstyle/data/DiffReport.java
@@ -147,10 +147,14 @@ public final class DiffReport {
                 : records.entrySet()) {
             final List<CheckstyleRecord> list = entry.getValue();
             for (CheckstyleRecord rec : list) {
-                statistics.addSeverityRecord(rec.getSeverity(),
-                        CheckstyleReportsParser.DIFF_REPORT_INDEX);
-                statistics.addModuleRecord(rec.getSource(),
-                        CheckstyleReportsParser.DIFF_REPORT_INDEX);
+                if (rec.getIndex() == CheckstyleReportsParser.BASE_REPORT_INDEX) {
+                    statistics.addSeverityRecordRemoved(rec.getSeverity());
+                    statistics.addModuleRecordRemoved(rec.getSource());
+                }
+                else {
+                    statistics.addSeverityRecordAdded(rec.getSeverity());
+                    statistics.addModuleRecordAdded(rec.getSource());
+                }
                 statistics.incrementUniqueMessageCount(rec.getIndex());
             }
         }
--- a/patch-diff-report-tool/src/main/java/com/github/checkstyle/data/Statistics.java
+++ b/patch-diff-report-tool/src/main/java/com/github/checkstyle/data/Statistics.java
@@ -19,10 +19,13 @@
 
 package com.github.checkstyle.data;
 
+import java.math.BigInteger;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import com.github.checkstyle.parser.CheckstyleReportsParser;
 
@@ -34,14 +37,24 @@ import com.github.checkstyle.parser.CheckstyleReportsParser;
 public class Statistics {
 
     /**
-     * Map storing severity numbers for difference.
+     * Map storing severity numbers for records removed.
      */
-    private Map<String, Integer> severityNumDiff = new HashMap<>();
+    private Map<String, BigInteger> severityNumDiffRemoved = new HashMap<>();
 
     /**
-     * Map storing module numbers for difference.
+     * Map storing severity numbers for records added.
      */
-    private Map<String, Integer> moduleNumDiff = new HashMap<>();
+    private Map<String, BigInteger> severityNumDiffAdded = new HashMap<>();
+
+    /**
+     * Map storing module numbers for records removed.
+     */
+    private Map<String, BigInteger> moduleNumDiffRemoved = new HashMap<>();
+
+    /**
+     * Map storing module numbers for records added.
+     */
+    private Map<String, BigInteger> moduleNumDiffAdded = new HashMap<>();
 
     /**
      * Number of files in difference.
@@ -51,12 +64,12 @@ public class Statistics {
     /**
      * Map storing severity numbers for base source.
      */
-    private Map<String, Integer> severityNumBase = new HashMap<>();
+    private Map<String, BigInteger> severityNumBase = new HashMap<>();
 
     /**
      * Map storing module numbers for base source.
      */
-    private Map<String, Integer> moduleNumBase = new HashMap<>();
+    private Map<String, BigInteger> moduleNumBase = new HashMap<>();
 
     /**
      * Number of files in the base source.
@@ -71,12 +84,12 @@ public class Statistics {
     /**
      * Map storing severity numbers for patch source.
      */
-    private Map<String, Integer> severityNumPatch = new HashMap<>();
+    private Map<String, BigInteger> severityNumPatch = new HashMap<>();
 
     /**
      * Map storing module numbers for patch source.
      */
-    private Map<String, Integer> moduleNumPatch = new HashMap<>();
+    private Map<String, BigInteger> moduleNumPatch = new HashMap<>();
 
     /**
      * Number of files in the patch source.
@@ -88,7 +101,15 @@ public class Statistics {
      */
     private int uniqueMessagesPatch;
 
-    public final Map<String, Integer> getSeverityNumDiff() {
+    /**
+     * Getter for number of records per severity for difference.
+     *
+     * @return number of records per severity.
+     */
+    public final Map<String, BigInteger> getSeverityNumDiff() {
+        final Map<String, BigInteger> severityNumDiff = new HashMap<>(severityNumDiffRemoved);
+        severityNumDiffAdded.forEach(
+            (key, value) -> severityNumDiff.merge(key, value, BigInteger::add));
         return severityNumDiff;
     }
 
@@ -97,15 +118,23 @@ public class Statistics {
      *
      * @return total number of severity records.
      */
-    public final int getTotalNumDiff() {
-        int totalSeverityNumber = 0;
-        for (Integer number : severityNumDiff.values()) {
-            totalSeverityNumber += number;
+    public final BigInteger getTotalNumDiff() {
+        BigInteger totalSeverityNumber = BigInteger.ZERO;
+        for (BigInteger number : getSeverityNumDiff().values()) {
+            totalSeverityNumber = totalSeverityNumber.add(number);
         }
         return totalSeverityNumber;
     }
 
-    public final Map<String, Integer> getModuleNumDiff() {
+    /**
+     * Getter for number of records per module for difference.
+     *
+     * @return number of records per module.
+     */
+    public final Map<String, BigInteger> getModuleNumDiff() {
+        final Map<String, BigInteger> moduleNumDiff = new HashMap<>(moduleNumDiffRemoved);
+        moduleNumDiffAdded.forEach(
+            (key, value) -> moduleNumDiff.merge(key, value, BigInteger::add));
         return moduleNumDiff;
     }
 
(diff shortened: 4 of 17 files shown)


romani:
@rnveach , 2 month without reply....
Do you have some concern on issue or implementation ? or just lack of time to review it?

@Vampire , please provide regression report, just to prove that change works, you can take any PR from main project for testing. I think lack of time test your PR, is a problem to allow merge.

patch-diff-report-tool/src/main/java/com/github/checkstyle/data/DiffReport.java
```diff
@@ -147,10 +147,17 @@ public void getDiffStatistics() {
                 : records.entrySet()) {
             final List<CheckstyleRecord> list = entry.getValue();
             for (CheckstyleRecord rec : list) {
-                statistics.addSeverityRecord(rec.getSeverity(),
-                        CheckstyleReportsParser.DIFF_REPORT_INDEX);
-                statistics.addModuleRecord(rec.getSource(),
-                        CheckstyleReportsParser.DIFF_REPORT_INDEX);
+                if (rec.getIndex() == CheckstyleReportsParser.BASE_REPORT_INDEX) {
```
rnveach: I don't see why we have to remove the `add` records and inline them here.
It seems to me this new logic should be in the `Statistics` class.

patch-diff-report-tool/src/main/java/com/github/checkstyle/data/DiffReport.java
```diff
@@ -147,10 +147,17 @@ public void getDiffStatistics() {
...
+                    statistics.addSeverityRecordRemoved(rec.getSeverity());
+                    statistics.addModuleRecordRemoved(rec.getSource());
+                }
+                else if (rec.getIndex() == CheckstyleReportsParser.PATCH_REPORT_INDEX) {
+                    statistics.addSeverityRecordAdded(rec.getSeverity());
+                    statistics.addModuleRecordAdded(rec.getSource());
+                }
+                else {
+                    throw new AssertionError();
```
rnveach: We should remove all these assertions as we are trying to get higher code coverage and there are no new tests to hit these.

patch-diff-report-tool/src/main/java/com/github/checkstyle/data/Statistics.java
```diff
@@ -88,27 +98,57 @@
...
     private int uniqueMessagesPatch;
 
+    /**
+     * Getter for number of records per severity for difference.
+     *
+     * @return number of records per severity.
+     */
     public final Map<String, Integer> getSeverityNumDiff() {
+        final Map<String, Integer> severityNumDiff = new HashMap<>(severityNumRemoved);
+        severityNumAdded.forEach((key, value) -> severityNumDiff.merge(key, value, Math::addExact));
```
rnveach: `Math::addExact` will throw an error if there is an overflow. I am not sure we really want this. There are alot more instances of this.

@romani ping.

patch-diff-report-tool/src/main/java/com/github/checkstyle/data/Statistics.java
```diff
@@ -34,14 +34,24 @@
...
      */
-    private Map<String, Integer> severityNumDiff = new HashMap<>();
+    private Map<String, Integer> severityNumRemoved = new HashMap<>();
 
     /**
-     * Map storing module numbers for difference.
+     * Map storing severity numbers for records added.
      */
-    private Map<String, Integer> moduleNumDiff = new HashMap<>();
+    private Map<String, Integer> severityNumAdded = new HashMap<>();
```
rnveach: IMO, we should `Diff` in the name since these are still for the difference line.

patch-diff-report-tool/src/main/resources/header.template
```diff
@@ -91,23 +91,23 @@
...
+							<td id="totalPatch" th:text="${statistics.totalNumPatch} + (${statistics.uniqueMessagesBase > 0 || statistics.uniqueMessagesPatch > 0} ? (' (' + (${statistics.uniqueMessagesBase > 0} ? |${statistics.uniqueMessagesBase} removed| + (${statistics.uniqueMessagesPatch > 0} ? ', ' : '') : '') + (${statistics.uniqueMessagesPatch > 0} ? |${statistics.uniqueMessagesPatch} added| : '') + ')') : '')"> totalNumPatch </td>
 							<th:block th:each="name : ${statistics.severityNames}">
-								<td th:id="${name} + @{SeverityNumPatch}" th:text="${statistics.severityNumPatch.get(name) != null} ? ${statistics.severityNumPatch.get(name)} : '0'"> infoNumBase </td>
+								<td th:id="${name} + @{SeverityNumPatch}" th:text="(${statistics.severityNumPatch.get(name)} ?: '0') + (${statistics.severityNumRemoved.get(name) || statistics.severityNumAdded.get(name)} ? (' (' + (${statistics.severityNumRemoved.get(name)} ? |${statistics.severityNumRemoved.get(name)} removed| + (${statistics.severityNumAdded.get(name)} ? ', ' : '') : '') + (${statistics.severityNumAdded.get(name)} ? |${statistics.severityNumAdded.get(name)} added| : '') + ')') : '')"> infoNumBase </td>
```
rnveach: Can we move these complex logics to some class so it can look nicer and easier to read?

patch-diff-report-tool/src/main/resources/header.template
```diff
@@ -124,19 +124,19 @@
...
 								<th:block th:each="name : ${statistics.moduleNames}">
-									<td th:id="${name} + @{ModuleNumBase}" th:text="${statistics.moduleNumBase.get(name) != null} ? ${statistics.moduleNumBase.get(name)} : '0'"> moduleNumBase </td>
+									<td th:id="${name} + @{ModuleNumBase}" th:text="${statistics.moduleNumBase.get(name)} ?: '0'"> moduleNumBase </td>
 								</th:block>
 							</tr>
 							<tr class="a">
 								<td>patch</td>
 								<th:block th:each="name : ${statistics.moduleNames}">
-									<td th:id="${name} + @{ModuleNumPatch}" th:text="${statistics.moduleNumPatch.get(name) != null} ? ${statistics.moduleNumPatch.get(name)} : '0'"> moduleNumPatch </td>
+									<td th:id="${name} + @{ModuleNumPatch}" th:text="(${statistics.moduleNumPatch.get(name)} ?: '0') + (${statistics.moduleNumRemoved.get(name) || statistics.moduleNumAdded.get(name)} ? (' (' + (${statistics.moduleNumRemoved.get(name)} ? |${statistics.moduleNumRemoved.get(name)} removed| + (${statistics.moduleNumAdded.get(name)} ? ', ' : '') : '') + (${statistics.moduleNumAdded.get(name)} ? |${statistics.moduleNumAdded.get(name)} added| : '') + ')') : '')"> moduleNumPatch </td>
```
rnveach: Same.

patch-diff-report-tool/src/main/resources/header.template
```diff
@@ -91,23 +91,23 @@
...
 							<th:block th:each="name : ${statistics.severityNames}">
-								<td th:id="${name} + @{SeverityNumPatch}" th:text="${statistics.severityNumPatch.get(name) != null} ? ${statistics.severityNumPatch.get(name)} : '0'"> infoNumBase </td>
+								<td th:id="${name} + @{SeverityNumPatch}" th:text="(${statistics.severityNumPatch.get(name)} ?: '0') + (${statistics.severityNumRemoved.get(name) || statistics.severityNumAdded.get(name)} ? (' (' + (${statistics.severityNumRemoved.get(name)} ? |${statistics.severityNumRemoved.get(name)} removed| + (${statistics.severityNumAdded.get(name)} ? ', ' : '') : '') + (${statistics.severityNumAdded.get(name)} ? |${statistics.severityNumAdded.get(name)} added| : '') + ')') : '')"> infoNumBase </td>
 							</th:block>
 						</tr>
 						<tr class="d">
-							<td>difference</td>
-							<td id="filesDiff" th:text=${statistics.fileNumDiff}> fileNumDiff </td>
-							<td id="totalDiff" th:text=${statistics.totalNumDiff}> totalNumDiff </td>
+							<td/>
```
rnveach: @romani What is your opinion on removing the word `difference`?
In issue some screenshots still had `difference`, some had `changes in violations`, and some had no text like this PR.

Vampire:
I'm on a road trip in USA without access to my computer. I'll be back end of July and then look into the open issues. But if you want to finish something before, be my guest.

patch-diff-report-tool/src/main/java/com/github/checkstyle/data/Statistics.java
romani: https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#addExact-int-int-
https://docs.oracle.com/javase/7/docs/api/java/lang/Integer.html#MAX_VALUE

@Vampire ,
Can we use BigInteger and standard arithmetic operation on them ? https://docs.oracle.com/javase/7/docs/api/java/math/BigInteger.html#add(java.math.BigInteger)

patch-diff-report-tool/src/main/resources/header.template
romani: all logic should be moved to java code. Template just show well prepared string values

patch-diff-report-tool/src/main/resources/header.template
romani: it is from column `Report Id` , so it is ok to remove, as we do not have such report ID, it is fake.
image without such word - https://github.com/checkstyle/contribution/issues/362#issuecomment-478247541
image with word - https://github.com/checkstyle/contribution/issues/362#issue-424254119

romani:
@Vampire , please continue this PR.
Unfortunately open and abandoned PRs , blocks others to continue fixing the issue. So we try to close long PRs to let somebody else to continue and finish functionality.

Lets say, if no reply/activity till end of August 2019, we consider that connection with you is lost and will close this PR.
Sorry that we did not replied to your PR quickly and there was 2 month delay on our side.

patch-diff-report-tool/src/main/java/com/github/checkstyle/data/DiffReport.java
Vampire: As it is quite some time I implemented this, I'm not sure anymore.
But I guess because `addSeverityRecord` is also used from other classes.
Before the change, the base and patch report numbers were just added together by calling add record for both of the other records with DIFF_REPORT_INDEX as index.
If you prefer I can of course instead add two new methods `addSeverityDiffRecord` and `addModuleDiffRecord` with the original record index as parameter and then do the `if` within statistics if you prefer.

patch-diff-report-tool/src/main/java/com/github/checkstyle/data/DiffReport.java
Vampire: done

patch-diff-report-tool/src/main/java/com/github/checkstyle/data/Statistics.java
Vampire: Well, I expected not to have more than Integer.MAX_VALUE findings and thought in case it does happen an exception is better than an overflow to MIN_VALUE. But we can of course use BigInteger.
done

patch-diff-report-tool/src/main/java/com/github/checkstyle/data/Statistics.java
Vampire: done

patch-diff-report-tool/src/main/resources/header.template
Vampire: done

patch-diff-report-tool/src/main/resources/header.template
Vampire: done

patch-diff-report-tool/src/main/resources/header.template
Vampire: This actually was what we agreed on in the issue, yes.
Also, it is not necessary anymore, as every column states what the number means now.

Vampire:
- updates are pushed
- example regression report at http://kautler.net/csc-regression-362/
- the updated test expectations also show the new format in different variations

romani:
link is opened as
```
Object not found!

The requested URL was not found on this server. If you entered the URL manually please check your spelling and try again.

If you think this is a server error, please contact the webmaster.
Error 404
kautler.net
Thu Aug 8 19:48:57 2019
Apache/2.2.9 (Debian) PHP/5.2.6-1+lenny16 with Suhosin-Patch mod_ssl/2.2.9 OpenSSL/0.9.8g 
```

romani:
@rnveach , please find time to finish review, it is better to resolved this PR sooner, while we remember a context of it.

Vampire:
Sorry, didn't upload the report yet, you were too fast, it is up now :-D

romani:
image

from my side probably is hard understanding of "6 changes in violations" I think it is better to be "6 diff items in violations". Because there were 3 changes, but we do not want to be that specific, we just show count of red/green lines in report below this statistic table.

@Vampire , can you regenerate report where few files are with changes ? You can take any PR we have now active and show how report become better. So it will max clear how it will looks like in real test case.

Vampire:
> from my side probably is hard understanding of "6 changes in violations" I think it is better to be "6 diff items in violations". Because there were 3 changes, but we do not want to be that specific, we just show count of red/green lines in report below this statistic table.

I'm not sure I understand what you mean.
The number in that line is like before my PR.
It was me who said the number is too meaningless and confusing.
It was @rnveach that suggested the worgin `x changes in violations`.
And it was you that picked that variant in the issue. :-D

If you want a different wording, I'm fine with that, whether it is `changes in violations` or `diff items in violations` I don't care, for me both strings have the exact same meaning and without more info are confusing. That's why we have now (after the PR) the counts for removed and added, that summed up make the number in the yellow line.

As you say `Because there were 3 changes`, I guess you are referring to the `Severity-warning` column. There the `6 changes in violations` in the yellow line is not a cound of the lower table simply, but only of the warning items. In this case it is the same as only warning items were different, but if you look at the `Severity-error` column, you see there is not 6 but 0 changes in violations.

Actually I personally like `changes in violations` more than `diff items in violations` but I have no too strong feelings about it. I just want to make sure you didn't loose context or misunderstood something. :-)

> @Vampire , can you regenerate report where few files are with changes ?

Sure, even if I'm not sure what that will change.

> You can take any PR we have now active and show how report become better. So it will max clear how it will looks like in real test case.

"Any PR we have now active" means which do you want and where do I get the config files for base and patch run? If you can provide some, I'll happily run with them, but I'm not going to investigate any active PR to understand what it is doing exactly just to construct two config files for the regression run. Especially as I don't really see the benefit of doing so actually. :-)

Vampire:
Nah, it's ok, I found my config for checkstyle/checkstyle#6466 and will run with that.
That is actually even the report that caused me to report the issue this PR is working on. :-D

romani:
> And it was you that picked that variant in the issue. :-D

and now I look at report, I do not understand what does it mean, it confuse me. Old naming was also a bit confused, but as started to improve, lets try a update wording a bit more, to make it firendly for new contributors. 
@rnveach , please review my proposal to slightly change wording. simple vote will resolve this.

I personally always look at exact report items, such statistis is low value for me during review. So I can accustom to any wording there.

Vampire:
Here the regression report for checkstyle/checkstyle#6466 with the new code: http://kautler.net/cs-regression-6466/

romani:
looks good for me

patch-diff-report-tool/config/import-control.xml
```diff
@@ -15,6 +15,9 @@
         <allow pkg="com.github.checkstyle.parser.StaxContentParser" />
         <allow pkg="com.github.checkstyle.PreparationUtils" />
         <allow class="org.thymeleaf.util.StringUtils" />
+        <file name="Statistics">
+            <allow class="java.math.BigInteger"/>
```
romani: please allow `java.math.BigInteger` new package in whole `data`.
testing tool should not be over strict on imports, we just enforced such Check to keep same style in project.

checkstyle-tester/diff.groovy
```diff
@@ -321,9 +321,9 @@ def getProjectsStatistic(diffDir) {
                 def indexHtmlFile = new File(fileObjf.absolutePath + '/index.html')
                 indexHtmlFile.eachLine {
                     line ->
-                        if (line.matches(".*totalDiff\">[0-9]+.*")) {
-                            def totalDiff = Integer.valueOf(line.substring(line.indexOf('>') + 1,
-                                    line.lastIndexOf('<')))
+                        def lineMatcher = line =~ /totalDiff">(?<totalDiff>[0-9]++)/
```
romani: I do not understand this syntax, can you make it simple (java style) ?
I do not want to force ocntributrs to have extra knowledge to read this code. 

It is written in groovy, only because there is no scripts in java.

patch-diff-report-tool/src/main/resources/header.template
romani: I am with removal

patch-diff-report-tool/src/main/java/com/github/checkstyle/data/DiffReport.java
romani: I am ok with either code.

rnveach:
@romani Feel free to rebase as there are conflicts.

romani:
merged manually.

PR #369, opened 2019-04-09
Issue #355: fix the referenced source files for local repositories that are not from "repositories"

Vampire:
A small oversight from implementing #355.
The paths to source files get replaced after report generation with a path in "repositories".
This is of course not valid if you work with "local" type repositories.


rnveach:
@Vampire CI is failing.
Failure is in codenarc reporting too many issues. Maybe this one put it over:
>     Violation: Rule=LineLength P=2 Line=99 Msg=[The line exceeds 120 characters. The line is 137 characters.] Src=[                    postProcessCheckstyleReport(targetDir, repoName, getOsSpecificPath(System.properties.'user.dir', reposDir, repoName))]

Vampire:
Yeah, you are probably right, sry.
Should be fixed now.

checkstyle-tester/launch.groovy
```diff
@@ -93,7 +93,10 @@ def generateCheckstyleReport(cliOptions) {
                     copyDir(getOsSpecificPath("$reposDir", "$repoName"), getOsSpecificPath("$srcDir", "$repoName"))
                 }
                 runMavenExecution(srcDir, excludes, checkstyleCfg, ignoreExceptions, checkstyleVersion, sevntuVersion)
-                postProcessCheckstyleReport(targetDir)
+                if (repoType != 'local') {
+                    repoUrl = getOsSpecificPath(System.properties.'user.dir', reposDir, repoName)
```
rnveach: I thought this fix was local type only. Why do we need changes for non-local?
Also why are we using the system user directory? All this is suppose to be local to the current repo.

Vampire: The fix is for local.
The original report as generated by checkstyle has paths like

`D:\Sourcecode\other\checkstyle-contribution\checkstyle-tester\src\main\java\android-launcher\src\com\android\launcher\AddAdapter.java`

The pre-PR code always replaces `checkstyle-contribution\checkstyle-tester\src\main\java` by `checkstyle-contribution\checkstyle-tester\repositories` and thus then has

`D:\Sourcecode\other\checkstyle-contribution\checkstyle-tester\repositories\android-launcher\src\com\android\launcher\AddAdapter.java`

The sense of the `local` type is to have the repo already present at any arbitrary place without copying it additionally to `repositories`, so the full path up to the repo name has to be replaced. For `local` the replacement value is the actually configured path `repoUrl`, for non-local repositories the replacement value is the respective full path inside the `repositories` directory.

romani:
@rnveach , please finish review.

rnveach:
@romani I will need to look at this in my local as I don't fully understand the fix and why this fix is for `local` but then all the new code says `not local`.

Another question is why the new changes to `repoUrl` have to be there when we use the original value of the variable above it.

romani:
@Vampire , please create issue on this to explain what is wrong and explain a reason of update.
We lost a context.

Unfortunately open and abandoned PRs make it hard to detect new meaningful and active PR. So we close all old PRs.

Lets say, if no reply/activity till end of August 2019, we consider that connection with you is lost and will close this PR.
Sorry that we did not replied to your PR quickly and there was 2 month delay on our side.

Vampire:
The issue is clearly described in the initial comment here:

> The paths to source files get replaced after report generation with a path in "repositories".
> This is of course not valid if you work with "local" type repositories.

> @romani I will need to look at this in my local as I don't fully understand the fix and why this fix is for local but then all the new code says not local.

Because for non-local it basically does it like before, for local not.

Let me explain further.
The checkstyle tester copies the sources from the original location to `src/main/java/repoName` and then generates the report.
This means in the report generated by CS are paths like
`/path/to/contribution/checkstyle-tester/src/main/java/repoName/what/Ever.java`.
Then `src/main/java` is wiped and the next repos sources are copied there.

The code before this PR always replaced
`getOsSpecificPath("checkstyle-tester", "src", "main", "java")` with
`getOsSpecificPath("checkstyle-tester", "repositories")`, so effectively
`checkstyle-tester/src/main/java` with
`checkstyle-tester/repositories`.

But with the `local` type this is not a valid assumption, as the repository can be anywhere on the computer.
So the code after this PR replaces for local repositories
`/path/to/contribution/checkstyle-tester/src/main/java/repoName` with
`/the/path/configured/for/the/local/repo`
and for non-local repositories
`/path/to/contribution/checkstyle-tester/src/main/java/repoName` with
`/path/to/contribution/checkstyle-tester/repositories/repoName`.

As not only the `src/main/java` part needs to replaced now, but the whole path, the actual replacement values changed for non-local too of course, or otherwise you would need to have two separate replacements, one for local, one for non-local, while this code duplication is not really necessary imho.

> Another question is why the new changes to repoUrl have to be there when we use the original value of the variable above it.

I don't understand that question.

Vampire:
You can also see the effect when you go to http://kautler.net/csc-regression-362/jEdit_clean/index.html and try to click any of the files in the last table. As the paths were not corrected correctly to the path of the local repository, the files were not actually collected and thus are not part of the final report, but only dead links.

checkstyle-tester/launch.groovy
romani: @rnveach ,

@Vampire mean that report is damaged by wrong links.
local mode does not do clone to "repositories" - https://github.com/checkstyle/contribution/commit/f7488e9d863f8abbdc11779fe7fbbd0f43c72c9c#diff-b56aba59072c339381ae8e62b6a145cdR89
so conversion of path should be done to NOT "repositories" folder, but to user defined path.
example: https://strkkk.github.io/4763/spotbugs/xref/checkstyle/contribution/checkstyle-tester/repositories/spotbugs/spotbugsTestCases/src/java/com/google/errorprone/bugpatterns/EmptyIfStatementNegativeCases.java.html#L18 (from https://strkkk.github.io/4763/)

the only problem I see is that reports that are based on "local" are not easy to share, as sources are not in "checkstyle-tester" folder. All links will be dead if report just copied to web.
We keep in repositories only required source files - https://github.com/strkkk/strkkk.github.io/tree/master/4763/spotbugs/xref/checkstyle/contribution/checkstyle-tester/repositories/spotbugs/spotbugsTestCases/src/java/com/google/errorprone/bugpatterns (not all files or original repo)
This probably should be mentioned in documentation, as attention note. So "local" mode is not designed for sharing in public web.

============

@Vampire , I am not comfortable with `System.properties.'user.dir'` can we keep code as it was before for mode `repoType != 'local` ? In this case postProcessCheckstyleReport will not do simply:
```
token: getOsSpecificPath(repoUrl),
value: getOsSpecificPath(repoDir)
```

PR #387, opened 2019-07-03
config: upgrade jgit to 5.4.0.201906121030-r in release notes

rnveach:


releasenotes-builder/src/main/java/com/github/checkstyle/github/NotesBuilder.java
```diff
@@ -254,7 +254,7 @@ private static boolean isIgnoredCommit(String commitMessage) {
      */
     private static ObjectId getActualRefObjectId(Repository repo, String ref) throws IOException {
         final ObjectId actualObjectId;
-        final Ref referenceObj = repo.getRef(ref);
+        final Ref referenceObj = repo.findRef(ref);
```
rnveach: I confirmed the javadocs are the same between these 2 versions, so this should be the right method.

romani:
Let's merge it, and see how release notes generated in checkstyle Travis

Issue #351, opened 2019-03-16
Please do shallow clones in checkstyle-tester when a tag or branch is to be tested


Vampire:
When using the checkstyle-tester it would be nice if it would do shallow clones with depth 1 for Git repos if possible.

If I for example do
`git clone https://github.com/google/guava`
the resulting directory is 283 MiB, if I instead do
`git clone --depth 1 --branch v25.0 https://github.com/google/guava`
the resulting directory is 43 MiB.

This saves significant time, bandwidth and disk-space.

romani:
Yes, our test tools have bunch of rooms to improve.
+1.

rnveach:
I'm not sure we can do this without some other breaking change.

> git clone --depth 1 --branch v25.0 https://github.com/google/guava

The field in the property file is `COMMIT_ID`. We cannot really know if it is a branch or a commit hash. Some of our repos use a hash.
I am not seeing a method on git clone to give it a hash, and `--branch` will not work with a hash.
If we use `--depth` by itself, the checkout command will fail saying it can't find the branch/hash.
> fatal: ambiguous argument 'v18.0': unknown revision or path not in the working tree.

Vampire:
> The field in the property file is `COMMIT_ID`. We cannot really know if it is a branch or a commit hash. Some of our repos use a hash.
> I am not seeing a method on git clone to give it a hash, and `--branch` will not work with a hash.
> If we use `--depth` by itself, the checkout command will fail saying it can't find the branch/hash.

Well, we could either change the format of the config file to either have two columns, one for ref names (branches or tags) that will do a shallow clone, one for other commit-ish values that will do a full clone and check out the specified commit-ish. (breaking of course)

Another way without a new column would be to support a prefix in the existing column, e. g. like `ref:v25.0` or `shallow:v25.0`, so that a shallow clone is done if the prefix is present and otherwise do a full clone and check out the specified commit-ish, no matter whether it is a ref or some other commit-ish. (non-breaking backwards compatible but needs explicit config for shallowness)

A third way would be to try a shallow clone and if it fails, do a full clone and check out the specified commit-ish. As the ref is directly searched in the remote repo to only clone that one commit, this will not need too much time in case of failure (non-breaking backwards compatible and automatically working without the need to specify shallowness manually)

> Even if we do do this, if we ever change the branch/commit, we would have to do a shallow fetch. It doesn't look like you can specify branch for a shallow fetch.

[...]

PR #391, opened 2019-07-04
Issue #351: add git shallow clone for all regression scripts

rnveach:
Issue #351

From what I have seen, `hg` has no equivalent.


rnveach:
> fatal: ambiguous argument 'v18.0': unknown revision or path not in the working tree.

PR #397, opened 2019-08-05
doc: suppression for javadoc parsing errors

pbludov:
From https://github.com/checkstyle/checkstyle/pull/6931#pullrequestreview-270506952

It is hard to read regression tests for a new javadoc check due to many parsing errors.
This PR shows how to suppress them.

### DIFF
--- a/checkstyle-tester/README.md
+++ b/checkstyle-tester/README.md
@@ -81,6 +81,32 @@ When the script finishes its work the following directory structure will be crea
 
 You will find *index.html* file in /reports/diff directory. The file represents summary diff report.
 
+## Javadoc regression:
+Many javadoc comments contain errors. To avoid report pollution with
+parsing errors, add suppression to the configuration:
+```
+<?xml version="1.0"?>
+<!DOCTYPE module PUBLIC
+  "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
+  "https://checkstyle.org/dtds/configuration_1_3.dtd">
+<module name="Checker">
+  <property name="haltOnException" value="false"/>
+  <module name="TreeWalker">
+    <module name="SummaryJavadoc" />
+
+    <!-- suppress javadoc parsing errors -->
+    <module name="SuppressionXpathSingleFilter">
+      <property name="message" value="Javadoc comment at column \d+ has parse error"/>
+    </module>
+  </module>
+  <!-- suppress java parsing errors -->
+  <module name="SuppressionSingleFilter">
+    <property name="message" value="Exception occurred while parsing"/>
+    <property name="checks" value="Checker"/>
+  </module>
+</module>
+```
+
 ATTENTION: 
 
 Administrators recommend you modify `projects-to-test-on.properties` and test as many projects as possible. Each project has its own unique style and it is common to find regression in 1 and not the others.
--- a/checkstyle-tester/my_check.xml
+++ b/checkstyle-tester/my_check.xml
@@ -26,11 +26,23 @@
          <module name="ThrowsCount">
              <property name="max" value="20000000"/>
          </module>
+         <!-- uncomment to suppress javadoc parsing errors
+         <module name="SuppressionXpathSingleFilter">
+            <property name="message" value="Javadoc comment at column \d+ has parse error"/>
+         </module>
+          -->
+
     </module>
 <!--
     <module name="SeverityMatchFilter">
         <property name="severity" value="warning"/>
         <property name="acceptOnMatch" value="false"/>
     </module>
+-->
+  <!-- uncomment to suppress java parsing errors
+  <module name="SuppressionSingleFilter">
+    <property name="message" value="Exception occurred while parsing"/>
+    <property name="checks" value="Checker"/>
+  </module>
 -->
 </module>


checkstyle-tester/README.md
```diff
@@ -121,6 +121,20 @@ With all that done, you can now call groovy. **baseBranch** will be your PR bran
 groovy diff.groovy --localGitRepo /home/johndoe/projects/checkstyle --baseBranch i111-my-fix --patchBranch i111-my-fix-mutation --config config.xml --listOfProjects projects-to-test-on.properties
 ```
 
+## Javadoc regression:
```
rnveach: I think this should be placed at the very bottom.
Where it is now isn't a good place anyways. `Attention` that appears afterwards is part `Checkstyle pitest regression` which you accidentally split up by placing it here.

What about naming this section more generic? We could add comments for other areas like `module-info.java` which also cause exceptions that pollute the report sometimes.

checkstyle-tester/README.md
```diff
@@ -121,6 +121,20 @@ With all that done, you can now call groovy. **baseBranch** will be your PR bran
...
+Many javadoc comments contain errors. To avoid report pollution with
+parsing errors, add suppression to the configuration:
+```
+<module name = "Checker">
+  <module name="SuppressionSingleFilter">
```
rnveach: Why checker filter over treewalker filter?

romani: better form of config is:
```
/var/tmp$ cat Test.java 
/** <a> bad html */
class Test {}

/var/tmp$ cat config.xml 
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
        "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
  <module name="TreeWalker">
    <module name="SummaryJavadoc" />

    <module name="SuppressionXpathSingleFilter">
      <property name="message" value="Javadoc comment at column \d+ has parse error"/>
    </module>
  </module>
</module>

/var/tmp$ java -jar checkstyle-8.23-all.jar -c config.xml Test.java 
Starting audit...
Audit done.
```

We should use Trwwwalker filter, as we clearly know from where we want to suppress violations.

checkstyle-tester/README.md
romani: agree. "Attention" block is for pitest chapter.

>What about naming this section more generic? 

I am not sure. I my mind it is better to have ready to use examples of config. Such README is targeted to new users, who do not want to spend his time to understand all details/nuances. Ones we have few sections like this, we will think about more general chapter.

rnveach: Ok, I am fine holding off on it then.

checkstyle-tester/README.md
romani: One more step to improve config:
```
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
  "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
  "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
  <property name="haltOnException" value="false"/>
  <module name="TreeWalker">
    <module name="SummaryJavadoc" />

    <module name="SuppressionXpathSingleFilter">
      <property name="message" value="Javadoc comment at column \d+ has parse error"/>
    </module>
  </module>
  <module name="SuppressionSingleFilter">
    <property name="message" value="Exception occurred while parsing"/>
  </module>
</module>
```

this will suppress failures like - https://pbludov.github.io/issue-6929/apache-ant/index.html#A2
that damage diff report too. I propose to place it in README and reuse.

@rnveach , please review. It was long discussion on this for a while with you on this, with no agreement. Lets try to discuss this one more time.

rnveach: I am fine with new config.
The only possible addition I could see is ensuring the filter matches the "rule" on the report so we don't accidentally suppress something else by mistake. I believe it is the source, not sure if that matches `checks` in the filter. https://checkstyle.sourceforge.io/config_filters.html#SuppressionSingleFilter_Properties

romani: good point,
I added `<property name="checks" value="Checker"/>` to it, it works, tested on local.

pbludov: all done.
All changes applied, the section moved  to right after the command line switches section.

checkstyle-tester/my_check.xml
```diff
@@ -26,11 +26,23 @@
          <module name="ThrowsCount">
              <property name="max" value="20000000"/>
          </module>
+         <!-- uncomment to suppress javadoc parsing errors
+         <module name="SuppressionXpathSingleFilter">
```
rnveach: I believe @romani wants this uncommented. This is why we try to make this as specific as possible.
@romani confirm

checkstyle-tester/README.md
rnveach: > Where it is now isn't a good place anyways. Attention that appears afterwards is part Checkstyle pitest regression which you accidentally split up by placing your new section here.

@pbludov this isn't done.

checkstyle-tester/my_check.xml
romani: good idea.
Filters are quick, so it will not be performance problems, but it will work as required out of the box.

romani:
code is changed and merged manually.

rnveach:
@romani code was merged but last point wasn't done. please wait on approval. https://github.com/checkstyle/contribution/pull/397#discussion_r310994525

rnveach:
Misclicked. Sorry.

checkstyle-tester/README.md
romani: @rnveach , I moved new section https://github.com/checkstyle/contribution/commit/d6228fc48e3b1f334585f552ecde3140293887cb

rnveach: Thanks.

PR #419, opened 2019-12-06
config: add spring-integration

strkkk:
https://github.com/checkstyle/checkstyle/pull/7315#issuecomment-562213799


strkkk:
@romani TC fails here too

Issue #422, opened 2019-12-14
diff report: wrap words, highlight target


romani:
right now 
long file names are not wrapped in Chrome (but works fine in FF)
![image](https://user-images.githubusercontent.com/812984/70855533-1dee6a80-1e81-11ea-8390-99110ad35640.png)
When you open certain link with defined target by anchor "#A11", it is hard to understand what actual line is it. 

should be:
![image](https://user-images.githubusercontent.com/812984/70855508-a91b3080-1e80-11ea-9c50-97796c0ee6f8.png)

rnveach:
Partial duplicate of https://github.com/checkstyle/contribution/issues/400 which already talks about adding text next to `#`.

rnveach:
fix was merged

PR #423, opened 2019-12-15
Issue #422: diff repot: wrap words, highlight target

romani:
Issue #422

result is show in issue.


patch-diff-report-tool/src/main/resources/maven-base.css
```diff
@@ -26,6 +26,9 @@ div.clear{
 div.clear hr{
   display: none;
 }
+:target {
```
rnveach: This should be `a:target` to show that you want to only highlight links.

romani: done, tested, works good

Issue #426, opened 2019-12-20
diff report should wrap all long content to looks better on mobile


romani:
https://gmshake.github.io/checkstyle_contribution/checkstyle-tester/diff_pr_7365_4/checkstyle/index.html
or
https://checkstyle.sourceforge.io/reports/7116-detect-throw/diff/apache-jsecurity/index.html

as diff reports should be reviewed even on the go
it is not very good for mobile :
![image](https://user-images.githubusercontent.com/812984/71226623-69de4c80-2291-11ea-9fa8-d0f245ec8b81.png)

------------

**Better form**:
![Screenshot_2019-12-19 checkstyle xml difference report](https://user-images.githubusercontent.com/812984/71226598-41eee900-2291-11ea-8fad-1667d40637eb.png)

rnveach:
@romani Does this close the issue?

romani:
yes, for now lets close it.

@jo2 , thanks a lot your work.

PR #427, opened 2019-12-20
Issue #426: wrap anywhere all long values in diff report

romani:
Issue #426

TODO: fix UTs

report generated by changes of this PR: https://romani.github.io/wrap/diff/guava/index.html

still problem: web site looks good in desktop browsers when size of window is small, BUT it stay the same in mobile browsers

### DIFF
--- a/patch-diff-report-tool/src/main/resources/maven-theme.css
+++ b/patch-diff-report-tool/src/main/resources/maven-theme.css
@@ -122,6 +122,9 @@ table.bodyTable td.c {
   background-color: #eee;
 }
 
+.wrap-anywhere {
+  word-break: break-all;
+}
 .source {
   border: 1px solid #999;
 }
--- a/patch-diff-report-tool/src/main/resources/content.template
+++ b/patch-diff-report-tool/src/main/resources/content.template
@@ -16,8 +16,8 @@
 							</th:block>
 								<td><a th:name="${anchor.getAnchor()}" th:href = "@{#} + ${anchor.getAnchor()}" th:text = "@{#} + ${anchor.getAnchor()}"> Anchor </a></td>
 								<td th:text = ${record.severity}> Severity </td>
-								<td th:text = ${record.simpleCuttedSourceName}> Rule </td>
-								<td th:utext = ${record.messageHtml}> Message </td>
+								<td class="wrap-anywhere" th:text = ${record.simpleCuttedSourceName}> Rule </td>
+								<td class="wrap-anywhere" th:utext = ${record.messageHtml}> Message </td>
 								<th:block th:switch="${record.xref}">
 									<td th:case = "null" th:text = ${record.line}> Line </td>
 									<td th:case = "*"><a th:href = "${record.xref} + @{#L} + ${record.line}" th:text = ${record.line}> Line </a></td>
--- a/patch-diff-report-tool/src/main/resources/header.template
+++ b/patch-diff-report-tool/src/main/resources/header.template
@@ -25,8 +25,8 @@
 							<th:block th:each="child : ${config.children}">
 								<th:block th:fragment="module" class="a">
 									<tr>
-										<td th:text=${child.simpleModuleName} class="c"> moduleName </td>
-										<td th:if="${child.hasBaseModule() and not child.hasIdenticalProperties}" class="b">
+										<td th:text=${child.simpleModuleName} class="c wrap-anywhere"> moduleName </td>
+										<td th:if="${child.hasBaseModule() and not child.hasIdenticalProperties}" class="b wrap-anywhere">
 											<th:block th:each="property : ${child.baseModuleProperties}">
 												<th:block th:text=${property.key}> propertyName </th:block>
 												<th:block th:each="token : ${property.value}">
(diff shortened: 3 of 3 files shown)


romani:
@jo2 , can you help me to figure out why mobile browsers do not follow `word-break: break-all;`

patch-diff-report-tool/src/main/resources/maven-theme.css
```diff
@@ -122,6 +122,9 @@ table.bodyTable td.c {
   background-color: #eee;
 }
 
+.wrap-anywhere {
```
rnveach: Why have a class name and just wrap all TDs/THs?

romani: If we apply such class to some html tag, it should work. We do have long values in headers of table (full name of module)

rnveach: If you change `.wrap-anywhere {` in the CSS to just `td, th {` it will apply to all TDs/THs in the document the CSS file applies to. Then you can just remove hardcoding the class name in the HTML.

romani: it will result in https://romani.github.io/wrap-all-td/diff/guava/index.html
wrap of all, even short columns, that are better to not break
Screenshot_2019-12-20 checkstyle xml difference report

rnveach: Ok, nevermind.

jo2:
@romani in your example, `word-break: break-all;` seems to work (in Firefox, Chrome has a totally different display, more on that later). `break-all` allows words to be separated however it fits best, dismissing any word separation guidelines like break on CamelCase or on syllabies and so on. Maybe using `word-break: break-word` optimizes it a bit:
image
vs: 
image
but it doesn't work for every table.

In some mobile browsers (Chrome, Chrome mobile and Samsung Internet are those i tested) tables are displayed like this: 
image

romani:
@jo2 , thanks a lot.
Do you have any recommendations on how change layout of such reports to make it work for mobile browsers ?

jo2:
I'll take a look on it on friday. Christmas break for now. Merry xmas!

jo2:
I have problems generating the output. Could you please add steps to do so?

romani:
examples of execution - https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#examples

you need to have two branches that behave differently. The most easy way is to make branch before fix of some issue and provide config that reference Check that was fixed.

Steps:
```
✔ ~/java/github/checkstyle/checkstyle [master|✔] 
15:20 $ git log -n 1
commit 5493363f5937aa136ab5f91ad49d202d937e66d9
Author: pbludov <pbludov@gmail.com>
Date:   Thu Dec 26 23:48:59 2019 +0300

    Issue #5777: Update AbbreviationAsWordInName to log DetailAST
```

```
✔ ~/java/github/checkstyle/contribution/checkstyle-tester [master|✚ 1] 
$ git diff
diff --git a/checkstyle-tester/my_check.xml b/checkstyle-tester/my_check.xml
index 2681927..30fd059 100644
--- a/checkstyle-tester/my_check.xml
+++ b/checkstyle-tester/my_check.xml
@@ -20,8 +20,7 @@
     <module name="TreeWalker">
          <!-- Example of checkstyle Check usage -->
          <!-- PLEASE CHANGE IT TO CHECK YOU ARE TESTING !!!! -->
-         <module name="ThrowsCount">
-             <property name="max" value="20000000"/>
+         <module name="AbbreviationAsWordInName">
          </module>
 
          <!-- Example of sevntu.checkstyle Check usage -->
✔ ~/java/github/checkstyle/contribution/checkstyle-tester [master|✚ 1] 
$ groovy diff.groovy --localGitRepo ~/java/github/checkstyle/checkstyle --baseBranch master --patchBranch master~1  --config my_check.xml --listOfProjects projects-to-test-on.properties


$ firefox reports/diff/index.html 
```

jo2:
The main problem is, that there is viewport information. Some browsers will scale each block in mobile view to fit the page. This causes the tables to seem so small. To fix this, add `<meta name="viewport" content="width=device-width, initial-scale=1">` in `patch-diff-report-tool/src/main/resources/header.template`. 
image

This can be improved by adding this css:
```
.bodyTable * {
    word-break: unset;
}

.wrapper {
    overflow-y: auto;
}
```
Also this has to be added in `header.template` (`<link rel="stylesheet" href="./css/normalize.css">`) and it has to be provided in `Main.java` (l. 247: `FilesystemUtils.exportResource("/filename.css", outputPath.resolve(CSS_FILEPATH).resolve("filename.css"));`). Also a `<div class="wrapper">` has to be put arround the tables in `header.template` and `content.template`. This results in scrollable tables:
image

@romani If you want, I can create a PR to your branch containing all this.

romani:
I always appreciate PRs.

You can create your branch base on master and do not bother with this/my PR.

> `Also a <div class="wrapper"> has to be put arround the tables `

I like no scrolling approach, as it keep reviewer focus on content without loosing position of review item.

jo2:
Added PR: https://github.com/checkstyle/contribution/pull/437

romani:
this PR is closed in favor of #437

PR #428, opened 2019-12-20
minor: patch-diff-report-tool: change font of h3 to small as file path is minor information

romani:
result:
https://romani.github.io/h3-small/diff/guava/index.html

from mobile:
Screenshot_20191220-171029

from laptop:
image

### DIFF
--- a/patch-diff-report-tool/src/main/resources/maven-theme.css
+++ b/patch-diff-report-tool/src/main/resources/maven-theme.css
@@ -51,7 +51,7 @@ h3 {
   color: #900;
   background-color: #eee;
   font-weight: normal;
-  font-size: large;
+  font-size: medium;
 }
 h4 {
   font-weight: normal;


rnveach:
It may be less important information, but I feel the text is too small to be read easily.

romani:
it is still mystery for me why fonts are so different on mobile and desktop.
I uploaded screenshot from mobile - still bigger than anything else.

Why do you need to read it ?
such path is from local host, so it is kind of useless, it is useful only when you need to copy whole path and do some CLI with such path on local (that is why full path is good).
When you read it in web it is useless, just some reference to name of file, and sometime I look at location in `main` or `resources`... but it is so rarely required.

romani:
ci failure is not related .... will be addressed in https://github.com/checkstyle/contribution/pull/429

rnveach:
> Why do you need to read it ?

If we don't need to read it, then why do we have it? If its displayed, I would like it to be readable so I'm not squinting, like for a first timer, to try and read it to see if it is important or not.

> such path is from local host, so it is kind of useless

Then something is wrong with utility, as it is all suppose to be repo path.

Example: http://rveach.no-ip.org/checkstyle/regression/reports/262/checkstyle/index.html

My report seems to be working fine, but I don't use groovy script.

patch-diff-report-tool/src/main/resources/maven-theme.css
```diff
@@ -51,7 +51,7 @@ h3 {
   color: #900;
   background-color: #eee;
   font-weight: normal;
-  font-size: large;
+  font-size: small;
```
rnveach: I recommend font size of normal.

romani:
https://www.w3schools.com/cssref/pr_font_font-size.asp - "medium" size.

rnveach:
@romani 
https://github.com/checkstyle/contribution/pull/428#issuecomment-568140348
> such path is from local host, so it is kind of useless

If this is still an issue and report doesn't localize the path then we should create an issue on it to fix it.

romani:
I still not sure how to do this in better way ...
let it be as it is now ... one day good idea will come up.

rnveach:
@romani As you pushed additional changes, please merge when you are good.

PR #437, opened 2019-12-30
Issue #426: add fix for chrome

jo2:
Issue #426

building on top of @romani's branch 426-wrap (#427). 

For improved readability I suggest making tables horizontal scrollable like we did for checkstyle documentation. 

Changes are uploaded to https://jo2.github.io/checkstyle-contribution/

### DIFF
--- a/patch-diff-report-tool/src/main/java/com/github/checkstyle/Main.java
+++ b/patch-diff-report-tool/src/main/java/com/github/checkstyle/Main.java
@@ -245,6 +245,8 @@ public final class Main {
                 outputPath.resolve(CSS_FILEPATH).resolve("maven-theme.css"));
         FilesystemUtils.exportResource("/maven-base.css",
                 outputPath.resolve(CSS_FILEPATH).resolve("maven-base.css"));
+        FilesystemUtils.exportResource("/site.css",
+            outputPath.resolve(CSS_FILEPATH).resolve("site.css"));
     }
 
     /**
--- a/patch-diff-report-tool/src/main/resources/content.template
+++ b/patch-diff-report-tool/src/main/resources/content.template
@@ -14,17 +14,17 @@
 								<tr th:case = "1" class="b">
 								<tr th:case = "*" class="a">
 							</th:block>
-								<td><a th:name="${anchor.getAnchor()}" th:href = "@{#} + ${anchor.getAnchor()}" th:text = "@{#} + ${anchor.getAnchor()}"> Anchor </a></td>
+								<td class="keepWidth"><a th:name="${anchor.getAnchor()}" th:href = "@{#} + ${anchor.getAnchor()}" th:text = "@{#} + ${anchor.getAnchor()}"> Anchor </a></td>
 								<td th:text = ${record.severity}> Severity </td>
 								<td th:text = ${record.simpleCuttedSourceName}> Rule </td>
 								<td th:utext = ${record.messageHtml}> Message </td>
 								<th:block th:switch="${record.xref}">
-									<td th:case = "null" th:text = ${record.line}> Line </td>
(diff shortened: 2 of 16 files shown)


romani:
```
$ git diff
diff --git a/checkstyle-tester/my_check.xml b/checkstyle-tester/my_check.xml
index 2681927..fcf843b 100644
--- a/checkstyle-tester/my_check.xml
+++ b/checkstyle-tester/my_check.xml
@@ -20,8 +20,8 @@
     <module name="TreeWalker">
          <!-- Example of checkstyle Check usage -->
          <!-- PLEASE CHANGE IT TO CHECK YOU ARE TESTING !!!! -->
-         <module name="ThrowsCount">
-             <property name="max" value="20000000"/>
+         <module name="AbbreviationAsWordInName">
+
          </module>
 
          <!-- Example of sevntu.checkstyle Check usage -->

✔ ~/java/github/checkstyle/contribution/checkstyle-tester [jo2/426-wrap-chrome-fix L|✚ 1] 
 $ groovy diff.groovy --localGitRepo ~/java/github/checkstyle/checkstyle \
     --baseBranch 5493363f5937aa136ab5f91ad49d202d937e66d9 \
     --patchBranch 5493363f5937aa136ab5f91ad49d202d937e66d9~1  \
     --config my_check.xml --listOfProjects projects-to-test-on.properties

```

romani:
@jo2 , really good ! but ....
it would be better to have violations column titles to be NOT wrapped.
and not wrapped values of column with anchors (first column) and "line"(before last) and "column" (last column).
I have feeling to make "Severity" to be not wrapped also, but probably it could take too much space.
Can we apply special css style (to print in tiny font) to such column when small screen is detected ? if not lets keep it wrapped, we will think about it later on.

jo2:
I can try applying a minimal width for those columns but we'll have to look how that works. Can you explain why horizontal scrolling is not an option?

romani:
> Can you explain why horizontal scrolling is not an option?

Probably I am over caution. But horizontal scrolling caused a lot of problems for me in past, so I almost fanatically try to eliminate it. That is why we have line length limit to 100 symbols - as it(github diff page) fits size of mobile in horizontal position.
So I apply the same rule to other areas that I do from mobile.
Review of diff report is same as review of code diff, reviewer need to be accurate and attentive and see not only single violation, but even few violations around. Some time we generate report for non single Check so violations in report can be  from different Checks (example https://sd1998.github.io/checkstyle-regression/Fix-7161/Hbase/index.html).
So reviewer need to see whole content of line.
As we do review most of time in our free time, such time is full of distractions, so no horizontal scrolling helps a lot in such cases.

That is why I would rather break words at any place (as it is still readable, even not pretty), to archive no horizontal scrolling.

if something in archiving of this is very complicated, lets discuss it, we will find compromise, we can improve step by step. Even in state of your PR, diff is much better than before.

jo2:
updated PR. The columns anchor, line and col are now forced to keep all characters in the line.

romani:
@jo2 , looks good ! https://romani.github.io/wrap-jo2/diff/guava/index.html

please fix CI, and lets proceed to ready to merge state.

romani:
CI is still not happy.
you can run https://github.com/checkstyle/contribution/blob/17cfea054adf9d1d4af86dd54301059e567e01eb/.ci/travis.sh#L15 to reproduce it on local

romani:
ones you pass CI, please squash all commits and make commit message reference issue number `Issue #426: ....`

jo2:
@romani thanks, will do so. I'll squash the commits afterwards

rnveach:
> Changes are uploaded to https://jo2.github.io/checkstyle-contribution/

Is this up to date so I can see changes?

romani:
@rnveach , please use https://romani.github.io/wrap-jo2/diff/guava/index.html , after my generation there were updates in UTs expected only.

PR #454, opened 2020-03-07
config: added JavadocMissingWhitespaceAfterAsteriskCheck

0blivious:
Add `JavadocMissingWhitespaceAfterAsteriskCheck` check to `checks-only-javadoc-error.xml`

reference: checkstyle/checkstyle#7775


rnveach:
Changes look right.

@romani CI should not pass here if we add a new module that is unreleased to either config. I can only assume like checkstyle, this file isn't being used in a CI here.

romani:
it might be used only by Checkstyle CI

PR #484, opened 2020-06-03
minor: Update checkstyle-tester README.md 

nmancus1:
minor: Update checkstyle-tester README.md to provide guide for Check and AST regression testing.

### DIFF
--- a/checkstyle-tester/README.md
+++ b/checkstyle-tester/README.md
@@ -224,6 +224,29 @@ Sevntu's current version must be referenced at https://github.com/checkstyle/con
 
 Finally, just run checkstyle-tester as described above.
 
+## Testing ANTLR Grammar Changes
+
+After modifying Checkstyle's ANTLR grammar, it is neccessary to compare how Checkstyle's checks are affected by your changes.  This is done by comparing the check violation output of your modified branch of Checkstyle against Checkstyle's master branch, and comparing the abstract syntax trees of both branches. 
+
+#### Check Regression Report
+
+To generate the check difference report, you must separate the modules and settings in the [`checkstyle_checks.xml`](https://github.com/checkstyle/checkstyle/blob/master/config/checkstyle_checks.xml) configuration file, found in the checkstyle repository, into separate configuration files (this helps avoid 'out of memory' errors).  The best way to separate the checks is to determine which typically produce the most violations, such as Indentation and WhiteSpace Checks, and put them each into their own configuration file, then separating the rest into files with roughly ten checks each.
+
+You may modify all the checks that depend on external files to use default settings.  **For each of the configuration files, you should use the [`my_check.xml`](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/my_check.xml) file as a base**, and add the checks from `checkstyle_checks.xml` to it. Then `diff.groovy` should be run on all projects in `projects-to-test-on.properties`, using the `diff.groovy` script once for each configuration file.  [See instructions above](https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#basic-difference-report) for "Basic Difference Report".
+
+#### ANTLR Regression Report
+
+ This report is generated using the [`launch_diff_antlr.sh`](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/launch_diff_antlr.sh) script.  This script generates a report based on the differences in the ASTs generated from your PR branch and Checkstyle's 'main' branch using projects that are selected (uncommented) in the [`projects-to-test-on.properties`](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/projects-to-test-on.properties) file. For the ANTLR regression report, we usually only want to see that changes to the Checkstyle project. To ensure that you test against any new inputs that you have created (unit test inputs, etc.), please make sure that you comment out all other projects, and add the following line to [`projects-to-test-on.properties`](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/projects-to-test-on.properties):
+ 
+ `my-checkstyle|git|https://github.com/<username>/checkstyle.git|<pr-branch>||`
+
+  Where `<username>` and `<pr-branch>` are your github username and the branch that your Checkstyle PR is based on, respectively. To use [`launch_diff_antlr.sh`](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/launch_diff_antlr.sh), you must modify the [`launch_diff_variables.sh`](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/launch_diff_variables.sh) file to reflect the location of the variables that  [`launch_diff_antlr.sh`](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/launch_diff_antlr.sh) uses to produce the report. Please note that `PULL_REMOTE` will need to be set to the name of the remote repository where your branch resides. If your branch is found at `origin/name-of-your-branch-here`, you will set `PULL_REMOTE=origin`. Then, run:
+  
+  `./launch_diff_antlr.sh name-of-your-branch-here`.
+
+*Note: if you are experiencing maven 'out of memory' errors from maven, see https://cwiki.apache.org/confluence/display/MAVEN/OutOfMemoryError*
+
+
 ## Troubleshooting
 
 To generate a report in debug mode, use the `MAVEN_OPTS` environment variable:


nmancus1:
I am converting this to draft status, I would like to add more information about testing with sevntu checkstyle.

checkstyle-tester/README.md
```diff
@@ -19,7 +19,7 @@ Depending on the type of external projects you wish to generate reports for, you
 
 **baseBranch** (b) - name of the base branch in local Checkstyle repository (optional, if asent, then the tool will use only patchBranch in case the tool mode is 'single', otherwise baseBrach will be set to 'master');
 
-**mode** (m) - the mode of the tool: 'single' or 'diff' (optional, default is 'diff'. Set this option to 'single' if your patch branch contains changes for any option that can't work on master/base branch. 
+**mode** (m) - the mode of the tool: 'single' or 'diff' (optional, default is 'diff'. Set this option to 'single' if your patch branch contains changes for any option that can't work on master/base branch.
```
rnveach, line 22: Make a separate PR for these fixes so we can merge right away.

checkstyle-tester/README.md
```diff
@@ -129,6 +129,12 @@ Example:
 </module>
 ```
 
+##### Modified Grammar Check Regression Report
+
+After modifying Checkstyle's ANTLR grammar, it is neccessary to compare how Checkstyle's checks are affected by your changes.  This is done by splitting the modules and settings in `checkstyle_checks.xml` configuration file, found in the checkstyle repository, into five separate configuration files (this helps avoid OOM).  You may modify all the modules that depend on external files to use default settings.  For each of the five configuration files, you should use the `my_checks.xml` file as a base, and add the modules from `checkstyle_checks.xml` to it. Then `diff.groovy` should be run on all projects in `projects-to-test-on.properties`, using the `diff.groovy` script once for each configuration file.  See instructions below for "Basic Difference Report".  If making changes to Checkstyle's grammar, also see "Testing Abstract Syntax Tree Regression" below.
```
rnveach, line 134: > into five separate configuration files 

The more checks we add the bigger the splits will be. I would try to guide on most optimally split method.

> See instructions below for "Basic Difference Report".
> see "Testing Abstract Syntax Tree Regression" below.

Just link to the instructions.

checkstyle-tester/README.md
```diff
@@ -219,11 +225,19 @@ First you must build sevntu checks:
...
 mvn  -Pno-validations clean install
-``` 
+```
 Sevntu's current version must be referenced at https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/pom.xml#L16 .
 
 Finally, just run checkstyle-tester as described above.
 
+## Abstract Syntax Tree Regression Report
+
+When making changes to Checkstyle's grammar, it is important to verify that there is no regression of the AST.  This is accomplished by using the `launch_diff_antlr.sh` script, and selecting one project at a time from `projects-to-test-on.properties`.  To use this script, first you must modify the `launch_diff_variables.sh` file to reflect the location of the variables that `launch_diff_antlr.sh` uses to produce the report. Please note that `PULL_REMOTE` will need to be set to the name of the remote repository where your branch resides. If your branch is found at `origin/name-of-your-branch-here`, you will set
```
rnveach, line 235: > and selecting one project at a time from `projects-to-test-on.properties`

Unless denoted otherwise, I believe Checkstyle should always be the 1 project. We pull in so many different sources from other places that we are unlike any other project.

nmancus1:
>I question why Grammar Check Regression and Abstract Syntax Tree Regression are 2 separate sections and so far apart.

I was thinking the same thing today.  Perhaps just make one section such as "Grammar Change Testing Procedure", and outline everything that must be done there.  What do you think of that?

rnveach:
It sounds good to me.

checkstyle-tester/README.md
nmancus1: Done.

checkstyle-tester/README.md
nmancus1: Fixed.

checkstyle-tester/README.md
nmancus1: Fixed, unless you think I should go into more detail.

checkstyle-tester/README.md
```diff
@@ -224,6 +224,22 @@ Sevntu's current version must be referenced at https://github.com/checkstyle/con
...
+
+#### Abstract Syntax Tree Regression Report
+
+When making changes to Checkstyle's grammar, it is important to verify that there is no regression of the AST.  This is accomplished by using the [`launch_diff_antlr.sh`](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/launch_diff_antlr.sh) script, and selecting (uncommenting) **only** the checkstyle project from [`projects-to-test-on.properties`](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/projects-to-test-on.properties).  To use this script, first you must modify the [`launch_diff_variables.sh`](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/launch_diff_variables.sh) file to reflect the location of the variables that  [`launch_diff_antlr.sh`](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/launch_diff_antlr.sh) uses to produce the report. Please note that `PULL_REMOTE` will need to be set to the name of the remote repository where your branch resides. If your branch is found at `origin/name-of-your-branch-here`, you will set
```
rnveach, line 237: I still don't see the reason for this separation. the antlr diff is still a diff on the grammar. If you just want to separate the 2 different types of runs, then I would start with the intro under the main header, `ANTLR Grammar Change Testing Procedure` and then go into the specifics of each run in the sub-header.

nmancus1: Fixed, I think.  Let me know if this is what you had in mind.

checkstyle-tester/README.md
```diff
@@ -224,6 +224,25 @@ Sevntu's current version must be referenced at https://github.com/checkstyle/con
...
+
+#### Abstract Syntax Tree Regression Report
+
+ This report is generated using the [`launch_diff_antlr.sh`](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/launch_diff_antlr.sh) script, and selecting (uncommenting) **only** the checkstyle project from [`projects-to-test-on.properties`](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/projects-to-test-on.properties).  To use this script, first you must modify the [`launch_diff_variables.sh`](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/launch_diff_variables.sh) file to reflect the location of the variables that  [`launch_diff_antlr.sh`](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/launch_diff_antlr.sh) uses to produce the report. Please note that `PULL_REMOTE` will need to be set to the name of the remote repository where your branch resides. If your branch is found at `origin/name-of-your-branch-here`, you will set
```
rnveach, line 237: >  **only** the checkstyle project

I would expand this to specifically say the `the checkstyle project sent as the PR`.

This is because master checkstyle won't have inputs for the new tree. So it will truly only be regression, but the PR is required to add new inputs to test the new features, so it will also show regression on the new changes. Some of which will just be exceptions turning into a tree for the full file, but it will atleast allow us to see the new tree and point to issues with it from the reports.

checkstyle-tester/README.md
```diff
@@ -224,6 +224,25 @@ Sevntu's current version must be referenced at https://github.com/checkstyle/con
 
 Finally, just run checkstyle-tester as described above.
 
+## ANTLR Grammar Change Testing Procedure
```
rnveach, line 227: => `Testing ANTLR Grammar Changes`

checkstyle-tester/README.md
```diff
@@ -224,6 +224,25 @@ Sevntu's current version must be referenced at https://github.com/checkstyle/con
...
...enerate the check difference report, you must separate the modules and settings in the [`checkstyle_checks.xml`](https://github.com/checkstyle/checkstyle/blob/master/config/checkstyle_checks.xml) configuration file, found in the checkstyle repository, into separate configuration files (this helps avoid OOM).  The best way to separate the modules is to determine which typically produce the most violations, such as Indentation and WhiteSpace Checks, and put them each into their own configuration file, then separating the rest into files with roughly ten modules each. You may modify all the modules that depend on external files to use default settings.  **For each of the configuration files, you should use the [`my_check.xml`](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/my_check.xml) file as a base**, and add the modules from `checkstyle_checks.xml` to it. Then `diff.groovy` should be run on all projects in `projects-to-test-on.properties`, using the `diff.groovy` script once for each configuration file.  [See instructions above](https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#basic-difference-report) for "Basic Difference Report".
```
rnveach, line 233: > Check Regression Report
> separate the modules and settings 

You call this check regression but you say modules in your description. If we are making ANTLR changes, you only need to regress on the checks and can ignore all FileSet modules. Only the checks will run against ANTLR. The base file sets from `my_check.xml` can stay as they support regression.

checkstyle-tester/README.md
```diff
@@ -224,6 +224,25 @@ Sevntu's current version must be referenced at https://github.com/checkstyle/con
...
+
+#### Abstract Syntax Tree Regression Report
```
rnveach, line 235: I am not sure about calling this AST since the name of the script is `antlr`.

checkstyle-tester/README.md
rnveach, line 233: > this helps avoid OOM
> Note: if you are experienceing maven OOM, see https://cwiki.apache.org/confluence/display/MAVEN/OutOfMemoryError

I feel it would be beneficial to specify what OOM is as many may not know it by that short name.

checkstyle-tester/README.md
```diff
@@ -224,6 +224,25 @@ Sevntu's current version must be referenced at https://github.com/checkstyle/con
...
+
+ This report is generated using the [`launch_diff_antlr.sh`](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/launch_diff_antlr.sh) script, and selecting (uncommenting) **only** the checkstyle project from [`projects-to-test-on.properties`](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/projects-to-test-on.properties).  To use this script, first you must modify the [`launch_diff_variables.sh`](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/launch_diff_variables.sh) file to reflect the location of the variables that  [`launch_diff_antlr.sh`](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/launch_diff_antlr.sh) uses to produce the report. Please note that `PULL_REMOTE` will need to be set to the name of the remote repository where your branch resides. If your branch is found at `origin/name-of-your-branch-here`, you will set
+`PULL_REMOTE=origin`. Then, run:
+```
+./launch_diff_antlr.sh name-of-your-branch-here
+```
+
+*Note: if you are experienceing maven OOM, see https://cwiki.apache.org/confluence/display/MAVEN/OutOfMemoryError*
```
rnveach, line 243: `experienceing` => `experiencing`

checkstyle-tester/README.md
nmancus1: I am confused; you're stating that we should amend `projects-to-test-on.properties` to checkout the contributor's PR branch?

checkstyle-tester/README.md
nmancus1: I was referring to xml 'modules', but I can see how that could be confused. Fixed.

checkstyle-tester/README.md
nmancus1: Fixed.

checkstyle-tester/README.md
nmancus1: Fixed.

checkstyle-tester/README.md
nmancus1: Fixed.

checkstyle-tester/README.md
nmancus1: Good catch, fixed.

checkstyle-tester/README.md
rnveach: Your report isn't providing `projects-to-test-on.properties` so I don't know which version of "checkstyle (2)" you used. Even specifying master or a local folder wouldn't give me this information either.

> you're stating that we should amend projects-to-test-on.properties to checkout the contributor's PR branch? 

Yes.

It is not that your report doesn't look correct, I am just saying it may be better to generate the report on the PR. It really comes down to the type of changes you are making and which version of the repo has the most to offer us in the regression report.

If you are adding brand new functionality that can't be parsed as is, Checkstyle master will not have any inputs to cover this case. Running regression on master will show old code is still handled the same way, but it won't show anything new. You will only get half the coverage. You won't see anything new.

If you are fixing existing functionality that is in a corner case, we may not have any inputs to cover this case in master. You won't see any differences in the tree to show the old problem is fixed. You will only see this if you run on the PR version of the repo.

PR checkstyle will always be master plus new input cases that cover more things. When you come down to it, PR may have more benefits to master. The only time this could not be the case is if we remove inputs from master, but that is usually rare.

nmancus1: This is an excellent point. I had not considered using the PR branch to test on, this is a great way to incorporate more information into the same report.  Thank you for explaining!

rnveach: No problem, I would add something like this to the read me changes.

One note I didn't hit on, when you add the PR version, you probably need to have it in a completely separate folder separated from the repo. Regression changes branches in the repo you give it. It will check out PR and master, so the files it runs on will change and may cause some issue with the report.

nmancus1: Ok, I've updated the README based on this conversation.  In order to simplify what we've discussed, I've just instructed the reader to comment out all projects and append a line to the file that will create a whole new folder and use their PR to run the AST regression test on. Let me know what else I can do here.

Issue #431, opened 2019-12-22
diff.groovy: add timestamp to main report next to commit hash


rnveach:
Add timestamp to main `diff.groovy` so reviewers can see when the commits being worked with were created. It will be easier than determining how old the commit used is in github's history.

rnveach:
Adding to report:
https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/diff.groovy#L297
https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/diff.groovy#L305

Generating value:
https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/diff.groovy#L194

We just need to add some minor additions to print the timestamp. We should get the timestamp from the commit being used, specifically the committed date.

romani:
Fix is merged

PR #488, opened 2020-06-30
Issue #431: add timestamp to diff.groovy main report

nmancus1:
Issue #431: add timestamp to diff.groovy main report
Instead of adding timestamp next to SHA,  I made a new line.  This will make the timestamp easier to read, and since I thought the time zone was a relevant addition, the line would have been too long if printed next to SHA.

Example report:
image

### DIFF
--- a/checkstyle-tester/diff.groovy
+++ b/checkstyle-tester/diff.groovy
@@ -192,7 +192,8 @@ def generateCheckstyleReport(cfg) {
     return new CheckstyleReportInfo(
         cfg.branch,
         getLastCommitSha(cfg.localGitRepo, cfg.branch),
-        getLastCommitMsg(cfg.localGitRepo, cfg.branch)
+        getLastCommitMsg(cfg.localGitRepo, cfg.branch),
+        getLastCommitTime(cfg.localGitRepo, cfg.branch)
     )
 }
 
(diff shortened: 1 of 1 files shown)


romani:
@rnveach , issue is created by you, please confirm that it is ok to merge.

rnveach:
Created by me but at your request. :)

checkstyle-tester/diff.groovy
```diff
@@ -343,6 +351,7 @@ def printReportInfoSection(summaryIndexHtml, checkstyleBaseReportInfo, checkstyl
     summaryIndexHtml << ('<br />')
     summaryIndexHtml << "Patch branch last commit message: \"$checkstylePatchReportInfo.commitMsg\""
     summaryIndexHtml << ('<br />')
+    summaryIndexHtml << "Patch branch last commit timestamp: \"$checkstylePatchReportInfo.commitTime\""
     summaryIndexHtml << ('<br />')
```
rnveach, line 355: Missing new `<br />`. Original code had 2 blank lines instead of 1.

rnveach:
Just curious, why is timezone different in screenshot between patch and base?

nmancus1:
>Just curious, why is timezone different in screenshot between patch and base?

That should be the time zone of the person that merged the PR (in this case, @strkkk)

rnveach:
Should we force timezone to be consistent?

nmancus1:
> Should we force timezone to be consistent between the 2? It will make reading easier.

If we do this, we should use GMT. I will update PR.

checkstyle-tester/diff.groovy
nmancus1: Fixed.

romani:
All reviewer is care is that:
- diff report is generated on same commit that he is reviewing 
- report is generated after commit is created.

Here is how we see it in github
Screen Shot 2020-06-30 at 12 55 14 PM

time local to me, so time in any other timezone is not very convenient to me, so it does not matter in what timze to show them.
timeze on master  commit also does not matter that much (let keep it as original).
So lets keep original timezones in all places as it was in first commit.

lets just print time of generation with contributor time zone to let me see report is after a date I see in github on commit.

nmancus1:
>So lets keep original timezones in all places as it was in first commit.

Done.

romani:
Please print time of report generation

nmancus1:
>Please print time of report generation

Example:
image

PR #494, opened 2020-07-12
patch-filter: include all files to analyze by Checkstyle

rdiachenko:


checkstyle-tester/pom.xml
```diff
@@ -74,6 +74,7 @@
 					<enableFilesSummary>false</enableFilesSummary>
 					<failsOnError>${checkstyle.failsOnError}</failsOnError>
 					<propertyExpansion>checkstyle.patchfilter.patch=${checkstyle.patchfilter.patch}</propertyExpansion>
+					<includes>**\/*</includes>
```
rnveach, line 77: What is this for? It should have been including all files to begin with. There is nothing here that says how issue was identified,

rdiachenko: @rnveach this won't go to master and was merged into patch-filter-jar branch which is a temporary branch to simplify report generation for the Patch filter project. This branch will eventually be removed with all changes in it.
But if you are still curious about the change, here's the reference - https://github.com/checkstyle/patch-filters/issues/31#issuecomment-657090755

Issue #344, opened 2019-01-13
patch-diff-report-tool: multi thread


rnveach:
Execution:
````
java -jar /home/ricky/opensource/contribution/patch-diff-report-tool/target/patch-diff-report-tool-0.1-SNAPSHOT-jar-with-dependencies.jar --baseReport /home/ricky/regression_reports/savemaster/openjdk10/results.xml --patchReport /home/ricky/regression_reports/savepull/openjdk10/results.xml --output /home/ricky/regression_reports/diff/openjdk10 --baseConfig /var/www/html/reports/32107f44546e135a34db874c19e7cfe4 --patchConfig /var/www/html/reports/32107f44546e135a34db874c19e7cfe4 --refFiles /home/ricky/regression_repositories/openjdk10
patch-diff-report-tool execution started.
XML parsing is started.
````
Config:
````
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
          "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">

<module name="Checker">
    <property name="charset" value="UTF-8"/>
    <property name="severity" value="warning"/>
    <property name="haltOnException" value="false"/>

    <module name="TreeWalker">

<module name="HiddenField">
  <property name="id" value="HiddenField1" />
</module>
<module name="HiddenField">
  <property name="id" value="HiddenField2" />
  <property name="ignoreConstructorParameter" value="true"/>
</module>
<module name="HiddenField">
  <property name="id" value="HiddenField3" />
  <property name="ignoreSetter" value="true"/>
</module>
<module name="HiddenField">
  <property name="id" value="HiddenField4" />
  <property name="setterCanReturnItsClass" value="true"/>
</module>

<module name="MagicNumber">
  <property name="id" value="MagicNumber1" />
</module>
<module name="MagicNumber">
  <property name="id" value="MagicNumber2" />
  <property name="ignoreHashCodeMethod" value="true" />
</module>
<module name="MagicNumber">
  <property name="id" value="MagicNumber3" />
  <property name="ignoreAnnotation" value="true" />
</module>
<module name="MagicNumber">
  <property name="id" value="MagicNumber4" />
  <property name="ignoreFieldDeclaration" value="true" />
</module>

<module name="ModifiedControlVariable">
  <property name="id" value="ModifiedControlVariable1" />
</module>
<module name="ModifiedControlVariable">
[...]

PR #514, opened 2020-10-23
Issue #344: add multi-thread support to CheckstyleReportsParser

rnveach:
Issue #344

I feel I have been sitting on this long enough. I have not run into any issues in all the times I have used it.

### DIFF
--- a/patch-diff-report-tool/src/main/java/com/github/checkstyle/data/DiffReport.java
+++ b/patch-diff-report-tool/src/main/java/com/github/checkstyle/data/DiffReport.java
@@ -24,7 +24,13 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.TreeMap;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 
 import com.github.checkstyle.parser.CheckstyleReportsParser;
 
@@ -37,6 +43,9 @@ import com.github.checkstyle.parser.CheckstyleReportsParser;
  *
  */
 public final class DiffReport {
+
+    /** Number of records to process at a time when looking for differences. */
+    private static final int SPLIT_SIZE = 100;
 
     /**
      * Container for parsed data,
@@ -61,6 +70,22 @@ public final class DiffReport {
 
     public Statistics getStatistics() {
         return statistics;
+    }
+
+    /**
+     * Utility to merge the patch related contents of the {@code other} class to the current.
+     *
+     * @param other The other class to merge.
+     */
+    public void mergePatch(DiffReport other) {
+        mergeRecords(other.records);
+        statistics.copyPatch(other.statistics);
+    }
+
+    private void mergeRecords(Map<String, List<CheckstyleRecord>> other) {
+        for (Entry<String, List<CheckstyleRecord>> item : other.entrySet()) {
+            addRecords(item.getValue(), item.getKey());
+        }
     }
 
     /**
@@ -103,16 +128,44 @@ public final class DiffReport {
      */
     private static List<CheckstyleRecord> produceDiff(
             List<CheckstyleRecord> list1, List<CheckstyleRecord> list2) {
+        final List<CheckstyleRecord> diff;
+        try {
+            diff = produceDiffEx(list1, list2);
+            diff.addAll(produceDiffEx(list2, list1));
+        }
+        catch (InterruptedException | ExecutionException ex) {
+            throw new IllegalStateException("Multi-threading failure reported", ex);
+        }
+
+        return diff;
+    }
+
+    private static List<CheckstyleRecord> produceDiffEx(
+            List<CheckstyleRecord> list1, List<CheckstyleRecord> list2)
+            throws InterruptedException, ExecutionException {
         final List<CheckstyleRecord> diff = new ArrayList<>();
-        for (CheckstyleRecord rec1 : list1) {
-            if (!isInList(list2, rec1)) {
-                diff.add(rec1);
+        if (list1.size() < SPLIT_SIZE) {
+            for (CheckstyleRecord rec1 : list1) {
+                if (!isInList(list2, rec1)) {
+                    diff.add(rec1);
+                }
             }
         }
-        for (CheckstyleRecord rec2 : list2) {
-            if (!isInList(list1, rec2)) {
-                diff.add(rec2);
+        else {
+            final ExecutorService executor =
+                Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors());
+            final List<Future<List<CheckstyleRecord>>> futures = new ArrayList<>();
+            final int size = list1.size();
+            for (int i = 0; i < size; i += SPLIT_SIZE) {
+                futures.add(executor.submit(new MultiThreadedDiff(list1, list2, i, Math.min(size, i
+                        + SPLIT_SIZE))));
             }
+
+            for (Future<List<CheckstyleRecord>> future : futures) {
+                diff.addAll(future.get());
+            }
+
+            executor.shutdown();
         }
         return diff;
     }
@@ -141,7 +194,7 @@ public final class DiffReport {
     /**
      * Generates statistical information and puts in in the accumulator.
      */
-    public void getDiffStatistics() {
+    public void generateDiffStatistics() {
         statistics.setFileNumDiff(records.size());
         for (Map.Entry<String, List<CheckstyleRecord>> entry
                 : records.entrySet()) {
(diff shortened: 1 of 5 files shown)


romani:
@rnveach , please write a summary of benefit to have MT mode. Is it only for extremely large reports ? Or it is benefitial on real life examples of use that we have currently in our PRs as regression report.

rnveach:
The benefit of MT is that the process will now split any work that was solely done in a single thread to be done simultaneously into separate threads, allowing the execution to be split among as many available threads open to the system for however big the reports and differences get.

Parsing of the 2 XML files, base and patch as represented by O(B + P) time (basically O(2n) time), is now split into 2 different threads allowing parsing time decreased to the smallest of either O(B) and O(P), which basically amounts to now O(n) time.

Finding differences on the violations in the XML files is where the most benefit is seen. Difference finding, which is comparing 2 lists and amounts to O(B * P) time (basically an O(n^2) complexity), is now split into as small as 100 record chunks to each thread that is available on the system. This haves a dramatic effect on time proportional to how many threads are on the system, theoretically amounting to O(B * P / T) time, where T is number of threads given to the process. The more threads available and the more records to process, the more time is saved with multi-threading. I think with all the savings, this may reduce the complexity to basically O(n log n) time. ( https://stackoverflow.com/questions/1592649/examples-of-algorithms-which-has-o1-on-log-n-and-olog-n-complexities )

> Is it only for extremely large reports ?

No. Anything with over 100 violations in any XML file will see improvements. The bigger, the more improvements seen.

>  is benefitial on real life examples of use that we have currently in our PRs as regression report.

Mine was a real life example as presented in issue. The more violations a produced in the final XML, the more savings will be generated. Pitest regression ( https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#checkstyle-pitest-regression ) multiplies the number of violations seen by 1 check and makes the number of violations bigger than normal. Indentation also produces possibly many violations just per file and will the be the single check to show the most improvement.

Remember improvements are seen based on the more violations in the XML file. This search routine is O(n^2) complexity.

romani:
I did not mean general ideas of why MT mode is better.

I meant, what is reduction of execution time for CI, for example for https://github.com/checkstyle/checkstyle/pull/8913 .

rnveach:
> I meant, what is reduction of execution time for CI

That is explained as I go into O(n) time complexity. It the notation used to describe how many comparisons a routine does. The less the time complexity, the less comparisons done, the faster the routine generally is.

O(1) is less time than O(n)
O(n) is less time than O(n log n)
O(n log n) is less time than O(n^2)
etc...

Main issue provides some numbers an actual run but I don't have the file to reproduce and provide more details from it, I can only explain the time reduction in terms of comparisons as there are many variables.

> for example for checkstyle/checkstyle#8913 .

Can you post to exact comment because I am not seeing anything.

romani:
I understand notation.
But if optimize process that takes 10 seconds to be 5 seconds, in Os it awesome but in real life it not that much , especially if whole other process takes 40 minutes. 

I curious how much quicker generation of report be after merge if this update. So I provided link to PR with configs.
I need to maintainability price of code in comparison with benefits of time reduction to get report ready.

patch-diff-report-tool/src/main/java/com/github/checkstyle/data/DiffReport.java
```diff
@@ -103,16 +128,44 @@ public void addRecords(List<CheckstyleRecord> newRecords,
...
-            if (!isInList(list2, rec1)) {
-                diff.add(rec1);
+        if (list1.size() < SPLIT_SIZE) {
+            for (CheckstyleRecord rec1 : list1) {
+                if (!isInList(list2, rec1)) {
+                    diff.add(rec1);
+                }
             }
         }
-        for (CheckstyleRecord rec2 : list2) {
```
pbludov, line 112: Why this loop is removed?

patch-diff-report-tool/src/main/java/com/github/checkstyle/data/DiffReport.java
```diff
@@ -103,16 +128,44 @@ public void addRecords(List<CheckstyleRecord> newRecords,
...
-            if (!isInList(list1, rec2)) {
-                diff.add(rec2);
+        else {
+            final ExecutorService executor =
+                Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors());
+            final List<Future<List<CheckstyleRecord>>> futures = new ArrayList<>();
+            final int size = list1.size();
```
pbludov, line 158: this looks as hardcoded
```java
    private static List<CheckstyleRecord> produceDiff(
            List<CheckstyleRecord> list1, List<CheckstyleRecord> list2) {
        return Stream.concat(
                StreamSupport.stream(list1.spliterator(), SPLIT_SIZE <= list1.size())
                    .filter(rec -> !isInList(list2, rec)),
                StreamSupport.stream(list2.spliterator(), SPLIT_SIZE <= list2.size())
                    .filter(rec -> !isInList(list1, rec))
            )
            .collect(Collectors.toList());
    }
```

patch-diff-report-tool/src/main/java/com/github/checkstyle/data/DiffReport.java
rnveach: This is the loop that amounts to O(N^2) complexity and is the real slowdown that needs multi-threading. The nested loop is inside `isInList`. It is now transferred to line 159 to split the list into pieces and send to the threads to complete. The rest of the loop is moved to line 270 which is inside the thread.

patch-diff-report-tool/src/main/java/com/github/checkstyle/data/DiffReport.java
rnveach: Is this multi-threaded? The whole issue is we need this multi-threaded to reduce processing time looping through the 2 lists.

pbludov: ` StreamSupport.stream(list1.spliterator(), SPLIT_SIZE <= list1.size())` is basically `list1.stream()` if the size is less then `SPLIT_SIZE`. Otherwise it is `list1.parallelStream()`

patch-diff-report-tool/src/main/java/com/github/checkstyle/data/DiffReport.java
pbludov: I don't get it at all. The code on line 159 will be executed only if the size of the list >= `SPLIT_SIZE`. Otherwise, the method will return only records that are in the list1 and are missing in the list2.

pbludov:
@rnveach what about such optimization:
```java
        List<CheckstyleRecord> shortest;
        List<CheckstyleRecord> longest;
        if (list1.size() < list2.size()) {
            shortest = list1;
            longest = list2;
        }
        else {
            shortest = list2;
            longest = list1;
        }

        // O(log(N))
        Set<CheckstyleRecord> recordSet =
                StreamSupport.stream(shortest.spliterator(), SPLIT_SIZE <= shortest.size()) // Do we need threading here?
            .collect(Collectors.toCollection(ConcurrentSkipListSet::new));

        // O(M * log(N)), M > N
        List<CheckstyleRecord> diff =
                StreamSupport.stream(longest.spliterator(), SPLIT_SIZE <= longest.size())
            .filter(rec -> !recordSet.remove(rec)) // Remove from both collections if matches
            .collect(Collectors.toCollection(ArrayList::new));

        diff.addAll(recordSet);
        return diff;
```
Could you run your benchmark on this code?

patch-diff-report-tool/src/main/java/com/github/checkstyle/data/DiffReport.java
rnveach: It sounds like you are missing https://github.com/checkstyle/contribution/blob/44f02903040bc7f16c834d26ed324e96b2c463a9/patch-diff-report-tool/src/main/java/com/github/checkstyle/data/DiffReport.java#L133-L134 which calls the same method with the lists swapped.

patch-diff-report-tool/src/main/java/com/github/checkstyle/data/DiffReport.java
rnveach: I am not familar with streams but I assume parallel is multi-threaded. I will compare and see how they stack against each other.

rnveach:
> what about such optimization:

I can add it to the list to benchmark but I don't understand the code to know what it is doing. Is this a replacement for `produceDiffEx` or `produceDiff`?

pbludov:
> Imagine both lists are of infinite size, it doesn't matter who is smaller. This is how I read `N` in this complexities since it isn't a direct number. It only matters what the final complexity is and if it can be reduced below `O(N^2)` or `O(N log N)`.

In this case, may be we should keep these lists *sorted*? It is easy to compare two sorted lists. Or even store them in a sorted collection, like
```java
    private Map<String, ConcurrentSkipListSet<CheckstyleRecord>> records
```
?

rnveach:
> In this case, may be we should keep these lists sorted?

It looks like we sorted the results after the diff. https://github.com/checkstyle/contribution/blob/44f02903040bc7f16c834d26ed324e96b2c463a9/patch-diff-report-tool/src/main/java/com/github/checkstyle/data/DiffReport.java#L113

If we are going to sort beforehand, wouldn't it be better to sort the records as they are added to the list?

romani:
@rnveach , please share comparison of executions, you might use Indentation Check to make sure that reports are huge in diff.

pbludov:
Please take a look to my PR #521

PR #521, opened 2020-10-31
Issue #344: add multi-thread support to CheckstyleReportsParser

pbludov:
Issue #344

@rnveach @romani @strkkk
Here is my attempt to speed up the diff report generation

* `CheckstyleRecord` now implements `Comparable`
* The method `specificEquals` now `Comparable::compareTo` as we need the order of records
* The list of records to process is pre-sorted (it has to be already sorted, so it does not take much time)
* The method `produceDiff` now is O(N) as both lists are sorted
* Difference generation is threaded to utilize more CPU and speed up the process
* `PositionOrderComparator` removed
* `produceDiff` now is a generic method and extracted to `DiffUtils` class
* `CliPaths` renamed to `CliOptions`
* New option for threading mode: SINGLE or MULTI

TODO:
- [x]  Do some regression for regression testing
- [x] Add an option to disable multithreading to easily diagnose any problems with threading that may occur in the future

### DIFF
--- a/patch-diff-report-tool/src/main/java/com/github/checkstyle/data/DiffReport.java
+++ b/patch-diff-report-tool/src/main/java/com/github/checkstyle/data/DiffReport.java
@@ -21,10 +21,10 @@ package com.github.checkstyle.data;
 
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
-import java.util.TreeMap;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentSkipListMap;
 
 import com.github.checkstyle.parser.CheckstyleReportsParser;
 
@@ -42,13 +42,34 @@ public final class DiffReport {
      * Container for parsed data,
      * note it is a TreeMap for memory keeping purposes.
      */
-    private Map<String, List<CheckstyleRecord>> records =
-            new TreeMap<>();
+    private final Map<String, List<CheckstyleRecord>> records =
+            new ConcurrentSkipListMap<>();
 
     /**
      * Container for statistical data.
      */
-    private Statistics statistics = new Statistics();
+    private final Statistics statistics = new Statistics();
+
+    /**
+     * Asynchronous tasks for difference calculation.
+     */
+    private final List<CompletableFuture<Void>> asyncTasks = new ArrayList<>();
+
+    /**
+     * Switch specifying if report generation should be done in a single-threaded
+     * or multi-threaded mode.
+     */
+    private final ThreadingMode threadingMode;
+
+    /**
+     * Creates instance of {@code DiffReport} class.
+     *
+     * @param threadingMode
+     *        type of threading mode to use.
+     */
+    public DiffReport(ThreadingMode threadingMode) {
+        this.threadingMode = threadingMode;
+    }
 
     /**
      * Getter for data container.
@@ -56,11 +77,50 @@ public final class DiffReport {
      * @return map containing parsed data.
      */
     public Map<String, List<CheckstyleRecord>> getRecords() {
-        return records;
+        return Collections.unmodifiableMap(records);
     }
 
     public Statistics getStatistics() {
         return statistics;
+    }
+
+    /**
+     * Adds new records to the diff report,
+     * when there are records with this filename, comparison
+     * between them and new record is performed and only difference is saved.
+     * If the report is generated in single-threaded mode, the difference will
+     * be calculated in-place. Otherwise, an asynchronous task will be scheduled.
+     *
+     * @param newRecords
+     *        a new records list.
+     * @param filename
+     *        name of a file which is a cause of records generation.
+     */
+    public void addRecords(List<CheckstyleRecord> newRecords, String filename) {
+        if (!newRecords.isEmpty()) {
+            if (threadingMode == ThreadingMode.SINGLE) {
+                addNewRecords(newRecords, filename);
+            }
+            else {
+                asyncTasks.add(CompletableFuture.runAsync(() -> {
+                    addNewRecords(newRecords, filename);
+                }));
+            }
+        }
+    }
+
+    /**
+     * Generates statistical information and puts in in the accumulator.
+     * This method will wait for completion of all asynchronous tasks.
+     */
+    public void getDiffStatistics() {
+        if (threadingMode == ThreadingMode.MULTI) {
+            joinAsyncTasksAll();
+        }
+        statistics.setFileNumDiff(records.size());
+        records.entrySet().stream()
+            .flatMap(entry -> entry.getValue().stream())
+            .forEach(this::addRecordStatistics);
     }
 
     /**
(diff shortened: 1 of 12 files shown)


pbludov:
Running diff on openjdk14 with modules
```xml
<module name="Indentation"/>
<module name="MagicNumber"/>
<module name="MissingCtor"/>
<module name="RequireThis"/>
<module name="ReturnCount">
  <property name="max" value="1"/>
  <property name="maxForVoid" value="0"/>
</module>
<module name="SimplifyBooleanExpression"/>
<module name="SimplifyBooleanReturn"/>
<module name="StringLiteralEquality"/>
<module name="VariableDeclarationUsageDistance"/>
<module name="AvoidStarImport"/>
<module name="AvoidStaticImport"/>
<module name="MissingJavadocMethod">
  <property name="allowMissingPropertyJavadoc" value="true"/>
</module>
<module name="MissingJavadocPackage"/>
<module name="MissingJavadocType"/>
<module name="JavaNCSS"/>
<module name="NPathComplexity"/>
```
produces checkstyle-result.xml with size 140M

## Diff generation timings (best result from 10 runs)

Master:
```
real    0m13,796s
user    0m16,818s
sys     0m0,251s
```

Patch:
```
real    0m4,518s
user    0m8,289s
sys     0m0,409s
```

romani:
Heh, from 13 second to 4 seconds, in worse case.
If we merge this , it will be only to please maintainers to put MT in code.

patch-diff-report-tool/src/main/java/com/github/checkstyle/data/DiffReport.java
```diff
@@ -76,6 +77,7 @@ public Statistics getStatistics() {
     public void addRecords(List<CheckstyleRecord> newRecords,
             String filename) {
         if (!newRecords.isEmpty()) {
+            newRecords.sort(new PositionOrderComparator());
```
pbludov, line 80: Why we are using `PositionOrderComparator` for sorting lines within a file?
This comparator uses only `line` and `column`, as a result there may be false positives for
```
line 100, col 100, messageA
line 100, col 100, messageB
```
versus
```
line 100, col 100, messageB
line 100, col 100, messageA
```
I'm going to use same `specificCompare` method here

romani: We did not review this code thoroughly,  it is just a code that works

pbludov: Done. Well, now i wonder why `CheckstyleRecord` does not implement `Comparable` but instead exposes a method used in sorting? This should be reviewed.

romani: Just fix it if you think it weird

pbludov: done

rnveach:
> If we merge this , it will be only to please maintainers to put MT in code.

It really depends on the differences. When I created the issue, I had a 1 hour run and MT reduced it to 20 minutes.

patch-diff-report-tool/src/main/java/com/github/checkstyle/data/SeverityLevel.java
```diff
@@ -0,0 +1,65 @@
...
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+////////////////////////////////////////////////////////////////////////////////
+
+package com.github.checkstyle.data;
+
+/**
+ * Severity level for a {@link CheckstyleRecord}.
+ * Used to compare two records in predefined order: "info", "warning", "error", all other.
+ */
+public enum SeverityLevel {
```
rnveach, line 26: Why not https://github.com/checkstyle/checkstyle/blob/2e8c277c181487d81231170ff4b959edc008114e/src/main/java/com/puppycrawl/tools/checkstyle/api/SeverityLevel.java#L34 ?

patch-diff-report-tool/src/main/java/com/github/checkstyle/data/DiffReport.java
```diff
@@ -76,6 +82,7 @@ public Statistics getStatistics() {
     public void addRecords(List<CheckstyleRecord> newRecords,
             String filename) {
         if (!newRecords.isEmpty()) {
+            Collections.sort(newRecords);
```
rnveach, line 85: As mentioned in previous PR, why not sort the records as they are added? Is that faster?

rnveach: Nevermind. https://stackoverflow.com/questions/168891/is-it-faster-to-sort-a-list-after-inserting-items-or-adding-them-to-a-sorted-lis

patch-diff-report-tool/src/main/java/com/github/checkstyle/data/DiffReport.java
```diff
@@ -76,111 +81,68 @@ public Statistics getStatistics() {
...
-                    statistics.addModuleRecordRemoved(rec.getSource());
-                }
-                else {
-                    statistics.addSeverityRecordAdded(rec.getSeverity());
-                    statistics.addModuleRecordAdded(rec.getSource());
-                }
-                statistics.incrementUniqueMessageCount(rec.getIndex());
-            }
-        }
+        records.entrySet().parallelStream()
```
pbludov, line 127: Parallelizing this loop gives 5% more speed for large diffs.

patch-diff-report-tool/src/main/java/com/github/checkstyle/data/SeverityLevel.java
pbludov: Diff report tool has no dependency on Checkstyle itself. This enum is to provide an order other then alphabetical to record severity. May be this enum is redundant and confusing. I'm going to review it.

patch-diff-report-tool/src/main/java/com/github/checkstyle/data/DiffReport.java
pbludov: upd: not in this PR. The class `Statistics` is not thread-safe.

pbludov:
The diff report is available at https://pbludov.github.io/

The first report is built with patch-diff-report-tool from `master`
The second report is from this branch in single-threaded mode
The third report is from this branch in multi-threaded mode

They are all equal.

patch-diff-report-tool/src/main/java/com/github/checkstyle/data/SeverityLevel.java
pbludov: done. @rnveach please review

rnveach:
To be updated as jobs run:

Files used: http://rveach.no-ip.org/checkstyle/regression/reports/294/ **(Warning XML is 169 megs)**
Command:
````
$ time java -jar ~/opensource/contribution/patch-diff-report-tool/target/patch-diff-report-tool-0.1-SNAPSHOT-jar-with-dependencies.jar --baseReport /var/www/html/reports/294/openjdk-pull.xml --patchReport /var/www/html/reports/294/openjdk-master.xml --output /tmp/test1 --baseConfig /var/www/html/reports/294/config.xml --patchConfig /var/www/html/reports/294/config.xml --refFiles /home/ricky/regression_repositories/openjdk10
````

TODO: more to be provided

rnveach:
@pbludov I do like your changes more. The thing is your changes aren't multi-threading. It is just pure optimization and so it all still runs single threaded. The quick run is just from 2 exact same files, I would like to see what is the time when everything is different. I assume it would then result in O(N^2) time since we have to go through each list twice, plus adding a sorting step which may result in longer times the more random the data gets.

I am fine if we want to keep this but still look into some way to multi-thread this.

rnveach:
@pbludov @romani Any thoughts?

pbludov:
> @pbludov @romani Any thoughts?

I also noticed that the only 1.2 cores are loaded. The main code is still single-threaded and has many ways for parallelization. But we can't go any further until we have a reliable method of regression testing.

pbludov:
@rnveach This takes much more time than I expected. Let's merge the optimization part and do multithreading later.

https://github.com/checkstyle/contribution/pull/533

Issue #524, opened 2020-11-15
migrate all openjdk repositories in projects-to-test-on.properties to github mirrors


nmancus1:
From https://github.com/checkstyle/checkstyle/pull/8987:

> > I have noticed this failure several times before, should we consider switching to one of the mirrors on github?
> 
> yes, better to use git version of code (github).

_Originally posted by @romani in https://github.com/checkstyle/checkstyle/issues/8987#issuecomment-727583722_

We have experienced numerous failures with the openjdk mercurial repositories, we should change all to GitHub mirrors so that they will be more reliable in the future.

OpenJDK7: https://github.com/openjdk-mirror/jdk7u-jdk
OpenJDK8: https://github.com/AdoptOpenJDK/openjdk-jdk8u
OpenJDK9: https://github.com/AdoptOpenJDK/openjdk-jdk9u
OpenJDK10: https://github.com/AdoptOpenJDK/openjdk-jdk10u
OpenJDK11: https://github.com/AdoptOpenJDK/openjdk-jdk11u

romani:
We can drop completely some old jdk that are easy to use

nmancus1:
> We can drop completely some old jdk that are easy to use

7, 8?

romani:
Yes, please, we can keep only latest.
As it will be less confusing for users to activate all or one.

Let's keep only one.

rnveach:
@romani Why are we removing old JDKs from test on, but not others like circle?

https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/projects-for-circle.properties#L8-L11

romani:
it is accumulation of trash .... they need to be removed. No value now.
Lets keep jdk14 only with clear list of excludes - it is more valuable.

nmancus1:
We can probably expand this issue to take care of all *.projects-*.properties., and if so, we can activate this after closing this issue: This project can be activated after this PR/ issue: https://github.com/checkstyle/checkstyle/blob/4c8ef953db33af72e1ee88645de826869f3fb186/.ci/shippable.sh#L59

romani:
fix is merged

PR #527, opened 2020-11-23
config: migrate all openjdk repositories in projects-to-test-on to github mirrors(#524)

nmancus1:
Issue #524: migrate all openjdk repositories in projects-to-test-on.properties to github mirrors

Openjdk 16 project has been added, and all others replaced.


nmancus1:
Before I mark this ready for review, we need to resolve two items:

- [ ] open an issue for `openjdk11-with-checks-nonjavadoc-error` if we decide to do that

- [ ] successful dummy Checkstyle PR using this file with "github, generate report"

nmancus1:
Regression on Openjdk10 is still suffering from a similar failure JXR failure that Openjdk11 and Openjdk 14.  I have switched to Openjdk9.

romani:
>Should we create a CI execution similar to openjdk14-with-checks-nonjavadoc-error for openjdk11?

No need. Let's keep single but well filtered.

rnveach:
@nmancus1 ping?

nmancus1:
@rnveach Sorry, I have finals this week and next, I was putting the removal of jxr and maven-checkstyle-plugin off until the end of next week.

nmancus1:
@rnveach Commit `92be3fb` is checked out as a branch per https://github.com/checkstyle/checkstyle/pull/9121#issuecomment-749725572, then built with `mvn clean install -Pno-validations`. Local maven repo contains the correct snapshot. The pom.xml in checkstyle-tester points to this version.  

`launch.groovy`:
Left is on the `master` branch of the contribution repo, and right is the `cleanup-projects` branch (my changes from https://github.com/checkstyle/contribution/pull/527):
image.  So, `launch.groovy` does not function as intended after removing the jxr plugin.

I will run `diff.groovy` next and see how that fairs.

rnveach:
@nmancus1 please split this PR between JXR removal and migrating mirrors.
https://github.com/checkstyle/contribution/issues/530 and https://github.com/checkstyle/contribution/issues/524 .

If you agree with plan at https://github.com/checkstyle/checkstyle/pull/9121#issuecomment-750724281 , gutting launch of HTML files is fine with me.

rnveach:
Blocked by https://github.com/checkstyle/contribution/issues/530

rnveach:
@nmancus1 Something happened with the PR, it contains the previous PR's changes too.

rnveach:
Also please tie this to #524 .

nmancus1:
> Something happened with the PR, it contains the previous PR's changes too.
Also please tie this to #524 in commit message.

Done, I thought I had fixed it before anyone noticed :)

checkstyle-tester/projects-to-test-on.properties
```diff
@@ -8,12 +8,7 @@
 #sevntu-checkstyle|git|https://github.com/sevntu-checkstyle/sevntu.checkstyle|master||
 #checkstyle-sonar|git|https://github.com/checkstyle/sonar-checkstyle|master||
 
-#openjdk7|hg|http://hg.openjdk.java.net/jdk7/jdk7/jdk/|||
-#openjdk8|hg|http://hg.openjdk.java.net/jdk8/jdk8/jdk/|default|
-#All details at checkstyle/checkstyle#3033: ModalDialogActivationTest till JDK-8166015 ; 'jxc/8073519/**' not compilable by design ; ', jhsdb/**' - checkstyle do not support unicode identifiers
-#openjdk9|hg|http://hg.openjdk.java.net/jdk9/jdk9/jdk/|default|**/test/java/awt/Focus/ModalDialogActivationTest/ModalDialogActivationTest.java,**/test/javax/xml/bind/jxc/8073519/**,**/test/sun/tools/jhsdb/**
-#openjdk10|hg|http://hg.openjdk.java.net/jdk10/jdk10/jdk/|default|**/test/javax/xml/bind/jxc/8073519/**,**/test/sun/tools/jhsdb/**
-#openjdk11|git|https://github.com/openjdk/jdk11u|master||
+#openjdk10|git|https://github.com/AdoptOpenJDK/openjdk-jdk10u|dev|**/test/javax/xml/bind/jxc/8073519/**,**/test/sun/tools/jhsdb/**
```
rnveach, line 11: Why is this 10 and not 11?
https://github.com/checkstyle/contribution/issues/524#issue-743347042

rnveach (commented):
Quick question..

rnveach (approved):

checkstyle-tester/projects-to-test-on.properties
nmancus1: There are files that cause an exception in Checkstyle in openjdk11; the list is nearly identical to the one found at https://github.com/checkstyle/checkstyle/blob/master/.ci/openjdk14-excluded.files.

rnveach: ok.

nmancus1: This way, we can simply tell a contributor to "uncomment all projects" when creating diff reports, with no caveats.

rnveach: Exceptions shouldn't matter to regression. Regression was built to report changes to exceptions. Like if a new exception came in or an old one went away. The only quirk to this was stack trace line numbers change and some don't want them to be displayed, regardless.

nmancus1:
projects file with new openjdk repo used in last diff report generation here: https://github.com/checkstyle/checkstyle/pull/9157

nmancus1:
After testing this numerous times on local and github action (as seen at https://github.com/checkstyle/checkstyle/pull/9157#issuecomment-755871941, https://github.com/checkstyle/checkstyle/pull/9157#issuecomment-755763863, and https://github.com/checkstyle/checkstyle/pull/9078#issuecomment-756093825) generation of Checkstyle report does not fail, but never completes ( I let it run for three days locally).  After diff'ing the two different repositories (https://github.com/AdoptOpenJDK/openjdk-jdk10u vs http://hg.openjdk.java.net/jdk10/jdk10/jdk/) there is a significant number of different files.

I have another idea; since OpenJDK10 does not recieve updates anymore, what if Checkstyle simply hosted it's own mirror of http://hg.openjdk.java.net/jdk10/jdk10/jdk/ for testing?

rnveach:
@nmancus1 
>  generation of Checkstyle report does not fail, but never completes ( I let it run for three days locally).

Use a custom version of Checkstyle, that prints the files as they are processing and you will find the file that is locking up regression.
Example:
https://github.com/rnveach/checkstyle/commits/more_audits
https://github.com/rnveach/checkstyle/commit/0807596764686c99d41cbe7238f868be95b48122

Most likely it is related to https://github.com/checkstyle/checkstyle/issues/8291 which was also found in regression.

> what if Checkstyle simply hosted it's own mirror of http://hg.openjdk.java.net/jdk10/jdk10/jdk/ for testing

If there is already an HTTP mirror, then we would need to know what is the goal of creating our own is. Any manual changes needed to the repo doesn't seem right since it just means we can't handle normal JDK files.

nmancus1:
> If there is already an HTTP mirror, then we would need to know what is the goal of creating our own is

My goal was to make sure that we have a reliable openjdk to test on when generating diff reports for check updates; I had many failures in report generation during last summer (and since then) using the openjdk mercurial repositories from the site being down.  I think it is important to test on them, since many of the issues I found when updating grammar were caught by the openjdk repos (and guava).

As it stands now, contributors are either testing on openjdk <= 9, or no openjdk at all because of confusion and previous failures.  The reason I chose openjdk10 is that it is the newest jdk we can test on with no excludes/ hacks; we can parse all files (except module-info.java).

rnveach:
I think you misunderstood my question. This PR already adds an HTTP version of JDK 10, and so will remove the HG. Why do we need our own version of HTTP (in checkstyle repo) versus the HTTP repo this PR gives us.

nmancus1:
@rnveach sorry, I did misunderstand.  The repo in this PR is the one that is taking forever to complete.  It is not identical to the mercurial repo ( to the tune of ~20,000 files).  I was proposing that we not use the one in this PR and instead make our own mirror of the mercurial repo.

rnveach:
>  The repo in this PR is the one that is taking forever to complete

Did you try running a custom checkstyle like i suggested to see if a specific file is causing it like the issue I pointed to?

nmancus1:
> Did you try running a custom checkstyle like i suggested to see if a specific file is causing it like the issue I pointed to? 

I did not try this yet.

> You can confirm the HG version runs to completion?

There are lots of examples, but here is one:
https://github.com/checkstyle/checkstyle/pull/8744

rnveach:
I tried out regression on this repository and regression halts on `/openjdk10test/test/langtools/tools/javac/lambda/speculative/NestedLambdaGenerics.java`

It is the exact same file and issue as https://github.com/checkstyle/checkstyle/issues/8291 and we cannot switch to this repo now.

rnveach (changes requested):
Can't be merged until the issue is resolved.

nmancus1:
This project can be activated after this PR/ issue: https://github.com/checkstyle/checkstyle/blob/4c8ef953db33af72e1ee88645de826869f3fb186/.ci/shippable.sh#L59

rnveach:
@nmancus1 It looks like connected issue is resolved. Maybe we could finish this now?

nmancus1:
@rnveach I have not been able to reproduce failure at https://github.com/checkstyle/checkstyle/pull/10911 locally at all. The way I see it, we can proceed a few different ways here:

1. Close this PR, leave issue for "someday"
2. Exclude the offending file. This means that we must allow exclusions in Checkstyle diff report generation.
3. Create a "Checkstyle" fork of the repo with offending file removed, just for diff report generation.

I would like to do (2). Please share your thoughts.

rnveach:
Is this the only file with an issue or are there more?

nmancus1:
> Is this the only file with an issue or are there more?

I think the only way to determine this definitively would be to allow exclusions in diff report generation and try it out in GitHub action CI. Since I cannot reproduce this failure locally, I can't guarantee that there isn't more files like this.

But, after thinking on this for awhile, we should really allow exclusions anyway, so that we can run regression in GitHub actions on the newest or at least more current jdk. We still really don't have any projects in the projects file that extensively use newer syntax.

nmancus1:
@rnveach to be clear, should I send a PR to main Checkstyle repo to allow exclusions in diff report generation?

rnveach:
Is main repo doing "launch" or diff?

The only way main repo can exclude files is to delete them after the cloning and even still I think the diff-side groovy script controls everything. It was not built for excluding files when doing regression. I was looking to understand how many files we would need to exclude to determine how we would handle this list, if it is just 1 file or many.

nmancus1:
> Is main repo doing "launch" or diff?

`launch.groovy` does not exist any longer.

> The only way main repo can exclude files is to delete them after the cloning and even still I think the diff-side groovy script controls everything. 

Support for exclusions was added to `diff.groovy` in #529

rnveach:
> Support for exclusions was added to diff.groovy in #529 

Except that support only exists for the launch side of things, not for difference reporting. Even if launch is gone, is functionality is in diff and still separate.

Since this is `projects-to-test-on` this means we are talking about the diff side so there is no exclusion functionality except if added to the config.

> Just because I can complete this task locally does not mean that it will succeed on github actions, the servers used are not spectacular:

Again, finishing locally will atleast tell us if we need to support a small list or a large list. If its a large list, we can't use launch's method of exclusion as it will become immense.

Please create new issue for this.

rnveach:
Use the property `allowExcludes`.
https://github.com/checkstyle/contribution/issues/580#issuecomment-969375012

nmancus1:
> Again, finishing locally will atleast tell us if we need to support a small list or a large list.

If we just use openjdk16 (which we should), we can look to https://github.com/checkstyle/checkstyle/blob/a7a3f7ac2766fddcc00b752f91e8c31cafdd61e6/.ci/openjdk16-excluded.files for the exact list of files to use

rnveach:
Yes, it makes more sense we combine this into 1 effort.

nmancus1:
Proof of successful parsing of new openjdk project here: https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/10a2890_2021173033/reports/diff/index.html

rnveach (approved):

checkstyle-tester/projects-to-test-on.properties
```diff
@@ -8,10 +8,8 @@
...
...ters.java,**/test/langtools/tools/javac/parser/SingleCommaAnnotationValueFail.java,**/test/langtools/tools/javac/switchextra/SwitchStatementBroken.java,**/test/langtools/tools/javac/switchextra/SwitchStatementBroken2.java,**/test/langtools/tools/javac/switchextra/SwitchArrowBrokenConstant.java,**/test/langtools/tools/javac/T8175198/AnnotationsAndFormalParamsTest.java,**/test/langtools/tools/javac/incompleteStatements/T8000484.java,**/test/langtools/tools/javac/T8181464/LambdaInAnnotationsCausesNPETest1.java,**/test/langtools/tools/javac/T8181464/LambdaInAnnotationsCausesNPETest2.java,**/test/langtools/tools/javac/8245153/T8245153.java,**/test/langtools/tools/javac/api/TestGetElementReferenceDataWithErrors.java,**/test/langtools/tools/javac/failover/FailOver15.java,**/test/langtools/tools/javac/failover/FailOver01.java,**/test/langtools/tools/javac/TryWithResources/PlainTry.java,**/test/langtools/tools/javac/var_implicit_lambda/VarInImplicitLambdaNegTest01.java,**/test/langtools/tools/javac/parser/8081769/T8081769.java,**/test/langtools/tools/javac/LabeledDeclaration.java,**/test/langtools/tools/javac/generics/6413682/T6413682.java,**/src/java.desktop/**,**/src/java.datatransfer/**|
```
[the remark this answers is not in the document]
romani: sure, this 10 years old quick-hack over idea how to do regression testing surprisingly still in service, but it is clear that there is bunch of need to make it different.

romani (approved):
ok to merge meanwhile it is upleasantly long line

Issue #534, opened 2020-12-27
Upgrade Ubuntu for Travis CI to 20.04


pbludov:
We are using an ancient version of Ubuntu for CI.
Although is a LTS release, libraries start to drop support for this build. Currently we have issues with markdownlint:
 
https://travis-ci.org/github/checkstyle/contribution/jobs/751662941#L244
```
ERROR:  Error installing mdl:

	The last version of chef-utils (>= 0) to support your Ruby & RubyGems was 16.6.14. Try installing it with `gem install chef-utils -v 16.6.14` and then running the current command again

	chef-utils requires Ruby version >= 2.6.0. The current ruby version is 2.5.3.105.
```

We should move on and use more recent distros for CI.

romani:
@pbludov , I temporarily disabled md verification to unblock CI

romani:
mdl problems:
```
checkstyle-tester/README.md:68: MD013 Line length

checkstyle-tester/README.md:69: MD013 Line length

checkstyle-tester/README.md:108: MD013 Line length

checkstyle-tester/README.md:117: MD013 Line length

checkstyle-tester/README.md:121: MD013 Line length

checkstyle-tester/README.md:124: MD013 Line length

checkstyle-tester/README.md:125: MD013 Line length

checkstyle-tester/README.md:102: MD032 Lists should be surrounded by blank lines

checkstyle-tester/README.md:114: MD038 Spaces inside code span elements

checkstyle-tester/README_GET_RAW_LINK.md:1: MD002 First header should be a top level header

checkstyle-tester/README_GET_RAW_LINK.md:3: MD022 Headers should be surrounded by blank lines

checkstyle-tester/README_GET_RAW_LINK.md:12: MD022 Headers should be surrounded by blank lines

checkstyle-tester/README_GET_RAW_LINK.md:1: MD041 First line in file should be a top level header

A detailed description of the rules is available at https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md
```

romani:
fix is merged

PR #535, opened 2020-12-27
Issue #534: upgrade Ubuntu to 20.04 for CI

pbludov:
Issue #534

I'm already using this distro on my PC, so no issues are expected.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,11 @@
 language: java
 sudo: false
 
-install:
-  - gem install mdl
+install: []
 
 matrix:
   fast_finish: true
   include:
-    - jdk: openjdk8
-      env:
-        - DESC="markdown lint"
-        - CMD="./.ci/travis.sh markdownlint"
-
     - jdk: openjdk8
       env:
         - DESC="releasenotes-builder"
@@ -52,7 +46,6 @@ addons:
   apt:
     packages:
       - xmlstarlet
-      - ruby
 
 script:
   - xmlstarlet ed --inplace -d "//mirrors" /home/travis/.m2/settings.xml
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -7,43 +8,43 @@ install:
 matrix:
   fast_finish: true
   include:
-    - jdk: openjdk8
+    - jdk: openjdk11
```
romani, line 11: I am not sure 😬 it good to drop jdk8 build validation.
Issue is about to please me linter.

romani (changes requested):
Item:

.travis.yml
pbludov: Ubuntu 20.04 in Travis CI has no support for openjdk8. Only openjdk 9+

romani: What if we activate azure ?
To make it like https://github.com/checkstyle/checkstyle/blob/74eb9219b6892f288cd293162666643b811ff5ff/azure-pipelines.yml#L117

pbludov: Yes, this should be azure pipiline. It is better to have common CI for all projects.

rnveach:
Until this is fixed, CI is of no use in this repo.

.travis.yml
romani: we have problems to share permission to azure project/organization ....
we can not come to agreement to [migrate to jdk11](https://github.com/checkstyle/checkstyle/issues/9146) in main repo ....

as this repo is not a library! jdk consideration are valid only for libraries. It is unlikely that we will use some jdk syntax feature so quickly.
I am completely ok to upgrade Travis to new OS version and use jdk11 for all builds in contribution repo.

Lets keep Azure as backup (second execution of the same), we need to comeback to this, but at least we will have config already.

pbludov:
closing for now, we can return to this issue after migration to java11 as the minimal version.

Issue #530, opened 2020-12-24
Remove JXR library from launch.groovy


rnveach:
As discussed in other threads this util has issues processing some Java and prevents us from moving forward with new java libraries. Remove JXR from launch will pretty much make launch useless as it won't generate any HTML files and it currently doesn't make use of patch-diff-report-util.

Removal of HTML is acceptable because of https://github.com/checkstyle/contribution/issues/529

Any code surrounding JXR in launch.groovy should be removed as it serves no purpose.

rnveach:
Fixed was merged

PR #536, opened 2020-12-29
Issue #530: Remove JXR library from launch.groovy

nmancus1:
Issue #530: Remove JXR library from launch.groovy
Ok, I wanted to provide as much proof as possible that this change has no ill effects; please let me know what else I can show. 

First, a simple example using two repos that I created (note that my `zsh` prompt shows the branch that I am currently on, as well as my notes).
One has non-compilable code that should cause exception (`broken-repo`) and one has compilable code (`not-broken`) that should( and does) allow for a `checkstyle-results.xml` to be generated: https://gist.github.com/nmancus1/d097e9654130329200849742ccf0971c

If anyone would like to verify the above, here are the projects file lines:
```
#broken|git|https://github.com/nmancus1/broken-repo.git|master||
#not-broken|git|https://github.com/nmancus1/not-broken.git|master||
```

Note at the end of the gist above, I diff the two `checkstyle-result.xml` files, successfully.
This shows that `launch.groovy` works as intended after removal of JXR and associated code in `launch.groovy`.
The only difference in the end result of `launch.groovy` is that the final report is not correctly cross-linked, as shown in https://github.com/checkstyle/contribution/pull/527#issuecomment-750363419. This is not a problem, because we only use launch.groovy to create `checkstyle-result.xml` files and validate that checkstyle can parse specified repositories.

Next, `diff.groovy` uses `launch.groovy` to create `checkstyle-result.xml` files that it compares across branches. So, as proof that `diff.groovy` still functions as intended, I ran a diff report with config and projects file from https://github.com/checkstyle/checkstyle/pull/9135, found here: https://nmancus1.github.io/remove-jxr-avoidescapes-diff/diff/index.html . ~~It looks like we have lost the links to source files with violations.  I am looking into this now, marked as draft.~~

### DIFF
--- a/checkstyle-tester/launch.groovy
+++ b/checkstyle-tester/launch.groovy
@@ -248,75 +248,11 @@ def runMavenExecution(srcDir, excludes, checkstyleConfig, ignoreExceptions, chec
 }
 
 def postProcessCheckstyleReport(targetDir, repoName, repoPath) {
-    def siteDir = getOsSpecificPath("$targetDir", "site")
-    println 'linking report to index.html'
-    new File(getOsSpecificPath("$siteDir", "index.html")).renameTo  getOsSpecificPath("$siteDir", "_index.html")
-    Files.createLink(Paths.get(getOsSpecificPath("$siteDir", "index.html")),
-        Paths.get(getOsSpecificPath("$siteDir", "checkstyle.html")))
-
-    removeNonReferencedXrefFiles(siteDir)
-    removeEmptyDirectories(new File(getOsSpecificPath("$siteDir", "xref")))
-
     new AntBuilder().replace(
         file: getOsSpecificPath("$targetDir", "checkstyle-result.xml"),
         token: new File(getOsSpecificPath("src", "main", "java", "$repoName")).absolutePath,
         value: getOsSpecificPath("$repoPath")
     )
-}
-
-def removeNonReferencedXrefFiles(siteDir) {
-    println 'Removing non referenced xref files in report ...'
-
-    def linesFromIndexHtml = Files.readAllLines(Paths.get("$siteDir/index.html"))
-    def filesReferencedInReport = getFilesReferencedInReport(linesFromIndexHtml)
-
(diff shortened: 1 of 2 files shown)


checkstyle-tester/launch.groovy
```diff
@@ -254,45 +253,13 @@ def postProcessCheckstyleReport(targetDir, repoName, repoPath) {
...
-                    && 'index.html' != fileName
-                    && 'overview-frame.html' != fileName
-                    && 'overview-summary.html' != fileName) {
-                    fileObj.delete()
-                }
-        }
-    }
-}
-
 def getFilesReferencedInReport(linesFromIndexHtml) {
```
rnveach, line 263: Can this be removed?

checkstyle-tester/launch.groovy
```diff
@@ -254,45 +253,13 @@ def postProcessCheckstyleReport(targetDir, repoName, repoPath) {
     Files.createLink(Paths.get(getOsSpecificPath("$siteDir", "index.html")),
         Paths.get(getOsSpecificPath("$siteDir", "checkstyle.html")))
```
rnveach, line 254: What about these items surrounding index/checkstyle HTML?
If HTML Java files are no longer supported, should we care if we have indexx/checkstyle HTML?

rnveach:
@romani I am good with these changes after my 2 comments, please review.

nmancus1:
Ok, links are fixed. Here is the output of a `diff` of the diff report found at https://github.com/checkstyle/checkstyle/pull/9135 and the one produced by this branch of contribution with my changes: https://gist.github.com/nmancus1/122cb586904d6e2972e806acee5c07cd

The only differences are times, directory paths, and the lack of empty xref directories.

checkstyle-tester/launch.groovy
nmancus1: Fixed.

checkstyle-tester/launch.groovy
nmancus1: Fixed.

rnveach (approved):

Issue #539, opened 2021-01-19
diff.groovy should allow configs to be placed in other directories


nmancus1:
As of https://github.com/checkstyle/contribution/commit/ed1f29895d9d0ea0eb20c6c7e1153961a89b5192, diff.groovy has not been able to generate crosslinked config files for the report summary outside of it's own directory. Noticed in numerous places, this limitation is causing failure for travis CI in https://github.com/checkstyle/regression-tool at https://travis-ci.org/github/checkstyle/regression-tool/jobs/755185633.  We have been using workarounds such as https://github.com/checkstyle/checkstyle/pull/9186/files#r560292164, but in the case of regression-tool, we would have to modify the tool to overcome this.

romani:
Fix is merged

PR #541, opened 2021-01-21
Issue #539: diff.groovy should allow configs in other directories

nmancus1:
Issue #539: diff.groovy should allow configs in other directories

Since this was only an issue when generating the html files to be crosslinked for the final report (not while generating the individual checkstyle reports), I simply copied all config files to be used to the working directory where the summary html is made. Instead of differentiating between different types of config files, this ends up just overwriting `projects-to-test-on.properties` with itself; I thought this would have no adverse affects and would make the code easier to read and maintain in the future.

Proof that this change fixes the final report regeneration mentioned in #539 : https://travis-ci.org/github/checkstyle/regression-tool/jobs/755600670

### DIFF
--- a/checkstyle-tester/diff.groovy
+++ b/checkstyle-tester/diff.groovy
@@ -1,5 +1,7 @@
 import static java.lang.System.err
+import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
 
+import java.nio.file.Files
 import java.nio.file.Paths
 import java.util.regex.Pattern
 
@@ -8,6 +10,7 @@ static void main(String[] args) {
     if (areValidCliOptions(cliOptions)) {
         def cfg = new Config(cliOptions)
         def configFilesList = [cfg.config, cfg.baseConfig, cfg.patchConfig, cfg.listOfProjects]
+        copyConfigFilesAndUpdatePaths(configFilesList)
 
         if (hasUnstagedChanges(cfg.localGitRepo)) {
             def exMsg = "Error: git repository ${cfg.localGitRepo.path} has unstaged changes!"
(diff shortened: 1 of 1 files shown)


checkstyle-tester/diff.groovy
```diff
@@ -153,6 +156,21 @@ def isExistingGitBranch(gitRepo, branchName) {
...
+    // Remove boolean value for single config in case of different base and patch config
+    configFilesList.removeIf { it instanceof Boolean };
+    // Remove dups in case of single config
+    configFilesList = configFilesList.unique();
+
+    for (filename in configFilesList) {
+        def sourceFile = new File(filename)
+        def checkstyleTesterDir = new File("").getCanonicalFile()
+        def destFile = new File("${checkstyleTesterDir.toPath()}/${sourceFile.getName()}")
+        Files.copy(sourceFile.toPath(), destFile.toPath(), REPLACE_EXISTING)
```
rnveach, line 169: Why do the files have to be copied? Why can't they be used from where they are?

This will cause extra files to appear in working directory, that will not be cleared away and show up as unstaged changes.

nmancus1: Due to the fact that config files are created in different branches, the file references get messed up, causing the NPE. I tried a few other implementations, one where we generate the html config files first, but it get pretty complicated. I had considered cleanup, but then thought that we don't cleanup /repositories or /reports, so why worry? I can certainly add cleanup to this.  I could create a "configHtml" directory that deleted after report generation.

nmancus1: We are usually running diff.groovy from ./ci-temp/ in the Checkstyle project, so file refs get messed up when we switch the branch from master to patch (or vice-versa).

rnveach: Is the issue just when you leave the config files in the same repo as the checkstyle repo that is being regressed on?

nmancus1: Yes, https://github.com/checkstyle/checkstyle/pull/9186/files#r560292164 is a good example.  I think this is ultimately the same problem in regression-tool, but I didn't dig into that repo too deep since it doesn't seem to be used anymore.

 My other thought to overcome this was to just create an option so that we can skip the final report generation, since when `diff.groovy` is used in this way, we don't need a final report anyway.

rnveach: That explains the issue. Truthfully it seems it should be given that you can't leave files in the repo being regressed on as the util will switch branches.

If @romani is fine with this then ok.

romani:
Let's keep it as is

PR #544, opened 2021-02-09
doc: rewrite diff report generation doc

strkkk:
After changes in https://github.com/checkstyle/checkstyle/pull/9258

Doc is rewritten to be clearer.
1. Github gist stuff was moved to separate page
2. Added description for new config line (it was missing)
3. Other enhancements and clarifications

Link to updated docs: https://github.com/strkkk/contribution/tree/diff_report_docs/checkstyle-tester#report-generation


checkstyle-tester/README.md
```diff
@@ -93,73 +93,70 @@ You can generate report in different ways:
...
-
-Alternative way is to commit file to some repository and branch, get its raw version and copy link.
-![Alt text](./screenshots/github_raw_button.png?raw=true "Repository raw button example")
+To generate report using github action, you need to add specific line(s) to your PR description.
+Please note:
+- Each item should be on new line.
+- There should be at least one config (items 1-3).
+- Placeholders should be replaced with links to ["raw" versions of files](./README_GET_RAW_LINK.md).
+
+Lines to add:
```
romani, line 102: unfortunately this section is hard to read and it is very scary for new contributor, our main goal is to not scary away new contributor.

lets keep this section as it is good for frequent contributors that need more options.

but what if we make at this level section like:
- Difference Report when you change existing Check
- Difference Report when you change existing Check (with custom projects list and label)
- Difference Report when you create new Check (with different base and patch config, with custom projects list and label)

so we will users exact link to use cases, without bothering to read whole manual and yes we can give them a link to manual section if still not clear.

romani (approved):
items for next PR:

checkstyle-tester/README.md
strkkk: there are too many cases with every option.
I can add some section to the top with name like "Quick start" and several most common options - new check, bugfix and new/different properties in checks without mention of labels or custom projects. These options can be listed after quick start section

PR #571, opened 2021-10-30
update all launch scripts

rnveach:
launch_diff :
* supports multiple configurations

all:
* executes ignored modules
* updated reports to be more similar to groovy's version


romani (approved):
I do not have time to verify that it works, lets see it real life

PR #583, opened 2021-12-04
checkstyle#10897: exclude file as failures become too annoying

romani:
checkstyle/checkstyle#10897


romani:
@rnveach , @pbludov , @strkkk , @nmancus1 , fyi ....

PR #595, opened 2022-01-22
Minor various fixes for diff-report-tool-util

rnveach:
Added missing documentation.
Fixed eclipse errors.
Renamed `CliPaths` to `CliOptions` since it does currently contain something that is not a path and I was thinking of expanding it for a new option. Also renamed `checkPaths` to `validate` since they aren't paths anymore. Split validate method between XML and Text for easier separation.


romani (approved):
Ok let it be

PR #608, opened 2022-05-25
Issue #607: Wercker service will be shutdown on October 31st, 2022

### LINKED ISSUE #607: Wercker service will be shutdown on October 31st, 2022
Child Issue: https://github.com/checkstyle/checkstyle/issues/11564

We need to migrate off Wercker and disable Wercker in all repos.

All the jobs which use `project-for-wercker.properties` should be moved to `project-to-test-on.properties`

Rahulkhinchi03:
Resolves #607


checkstyle-tester/projects-to-test-on.properties
```diff
@@ -52,3 +52,32 @@ guava|git|https://github.com/google/guava|v28.2||
...
+
+#hibernate-orm|git|https://github.com/hibernate/hibernate-orm|4.2.19.Final|**/hibernate-orm/documentation/**/*
+
+#spotbugs|git|https://github.com/spotbugs/spotbugs|3.1.2||
+
+#spring-framework|git|https://github.com/spring-projects/spring-framework|v4.1.6.RELEASE||
+
+#Hbase|git|https://github.com/apache/hbase|1.1.0.1||
+
+#pmd|git|https://github.com/pmd/pmd|pmd_releases/6.21.0|**/pmd/pmd-java/src/test/**/*,**/pmd/cpd/files/*
```
rnveach, line 70: As an example with your changes, PMD is listed twice. We need each item to be unique or rename the first column so it is unique.

I don't see why we would have duplicate clone URLs for this issue.

rnveach (changes requested):

rnveach:
@romani ping on TC failures here.

checkstyle-tester/projects-to-test-on.properties
Rahulkhinchi03: Understood. Changes Done

rnveach (approved):
@Rahulkhinchi03 You confirmed everything else is the same, including the exclusions?

Rahulkhinchi03:
> @Rahulkhinchi03 You confirmed everything else is the same, including the exclusions?

@rnveach Yes, I have cross-checked it.

romani (approved):
Ok to merge, TC is not related.
We need to fix it in separate PR.

PR #613, opened 2022-07-01
minor: create latest projects file for regression testing

nick-mancuso:
This is just a copy/paste/modify of the existing projects file, where specific commits and tags have been removed to allow us to test on the latest version of the project.

See https://github.com/checkstyle/checkstyle/pull/11805#issue-1291342040 and https://github.com/checkstyle/checkstyle/pull/11805#issuecomment-1172370264  for proof that Checkstyle can complete execution with these projects/ properties.

### DIFF
--- a/checkstyle-tester/latest-projects-to-test-on.properties
+++ b/checkstyle-tester/latest-projects-to-test-on.properties
@@ -1,0 +1,27 @@
+# List of GIT repositories to clone / pull for checking with Checkstyle
+# File format: REPO_NAME|[local|git|hg]|URL|[COMMIT_ID]|[EXCLUDE FOLDERS]
+# Please note that bash comments works in this file
+
+guava-latest|git|https://github.com/google/guava|||
+
+spotbugs-latest|git|https://github.com/spotbugs/spotbugs|||
+pmd-latest|git|https://github.com/pmd/pmd||
+spoon-latest|git|https://github.com/INRIA/spoon.git||
+lombok-ast-latest|git|https://github.com/rzwitserloot/lombok.ast||
+spring-framework-latest|git|https://github.com/spring-projects/spring-framework||
+hibernate-orm-latest|git|https://github.com/hibernate/hibernate-orm||
+elasticsearch-latest|git|https://github.com/elastic/elasticsearch|||
+java-design-patterns-latest|git|https://github.com/iluwatar/java-design-patterns|||
+MaterialDesignLibrary-latest|git|https://github.com/navasmdc/MaterialDesignLibrary|||
+Hbase-latest|git|https://github.com/apache/hbase|||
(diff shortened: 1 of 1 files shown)


checkstyle-tester/latest-projects-to-test-on.properties
```diff
@@ -0,0 +1,28 @@
+# List of GIT repositories to clone / pull for checking with Checkstyle
+# File format: REPO_NAME|[local|git|hg]|URL|[COMMIT_ID]|[EXCLUDE FOLDERS]
+# Please note that bash comments works in this file
+
+guava|git|https://github.com/google/guava|v28.2||
```
rnveach, line 5: Why don't we just remove the specific versions from all of them and have them always be the latest?

Do we want to add this as a feature to the groovy script to ignore commit/tag?

nick-mancuso: > Guava still specifies a specific version.

Mistake, fixed.

> Do we want to add this as a feature to the groovy script to ignore commit/tag?

I would rather not do this, for a few reasons:
1. I would rather get by with current groovy script and put future effort into a whole new system
2. This still does not help us for check regression report in github CI, which is probably how 98% of reports are generated these days. 

We could add some options to github action that would translate to groovy script execution, but I would prefer to not complicate that more either.

nick-mancuso: Generating new reports at  https://github.com/checkstyle/checkstyle/pull/11805, due to missed tag at https://github.com/checkstyle/contribution/pull/613#discussion_r912063874

nick-mancuso:
Generating new reports at  https://github.com/checkstyle/checkstyle/pull/11805, due to missed tag at https://github.com/checkstyle/contribution/pull/613#discussion_r912063874

checkstyle-tester/latest-projects-to-test-on.properties
```diff
@@ -0,0 +1,28 @@
+# List of GIT repositories to clone / pull for checking with Checkstyle
+# File format: REPO_NAME|[local|git|hg]|URL|[COMMIT_ID]|[EXCLUDE FOLDERS]
+# Please note that bash comments works in this file
+
+guava|git|https://github.com/google/guava|||
+
+spotbugs|git|https://github.com/spotbugs/spotbugs|||
```
nick-mancuso, line 7: Let’s use `guava-latest` naming convention for all projects here to allow for easy copy paste and no conflicts on local.

nick-mancuso: Done

romani:
I am not sure that latest sources is stable execution, but if we want to play with it - ok.
As option it is good.

romani (approved):

rnveach (approved):
I am still not liking creating a new file for every off shot need.

PR #635, opened 2022-08-19
Issue #634: satelite projects should use latest checkstyle as soon at is appears in maven repo

### LINKED ISSUE #634: satelite projects should use latest checkstyle as soon at is appears in maven repo
Right now we do bump a version in a few projects https://github.com/checkstyle/checkstyle/wiki/How-to-make-a-release#update-for-released-version-reference right after release.

Let's find a way to define in pom.xml the latest snapshot and most recent version. And the latest released version when we define dependencies. To let us skip such updates at all.

CI of such projects should be able to get the latest version from https://repo1.maven.org/maven2/com/puppycrawl/tools/checkstyle/maven-metadata.xml and run projects with such version.

We can have in all such repos, a GitHub action that runs once a day check https://repo1.maven.org/maven2/com/puppycrawl/tools/checkstyle/maven-metadata.xml, put such values in required files, and send PRs with changes (PR is necessary as we need notice is any problems happens if we introduce breaking changes)

TASK:

- [ ] Version Bump Script in `releasenotes-builder/pom.xml`
- [ ]  Version Bump Script in `patch-diff-report-tool/pom.xml`
- [ ]  GitHub Action to run both the scripts.

Rahulkhinchi03:
Partially Fixes #634

### DIFF
--- a/.ci/bump-cs-version-in-releasenotes-builder.sh
+++ b/.ci/bump-cs-version-in-releasenotes-builder.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+set -e
+
+cd releasenotes-builder
+mvn versions:set-property -Dproperty=checkstyle.version -DnewVersion="$1"
+echo "Version updated to $1"


.ci/bump-releasenotes-builder-version.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
romani, line 1: name of script should be `bump-cs-version-in-releasenotes-builder.sh`
please apply the same naming to all other shell files of such type that you are introducing.

.ci/bump-releasenotes-builder-version.sh
```diff
@@ -0,0 +1,24 @@
...
+
+MVN_CURRENT_VERSION=$(grep "^        <checkstyle.version>.*</checkstyle.version>$" "$POM_FILE" | awk -F'[><]' '{print $3}')
+echo MVN_CURRENT_VERSION="$MVN_CURRENT_VERSION"
+
+echo "Checking Version for: "$POM_FILE
+if [ "$MVN_CURRENT_VERSION" != "$LATEST_VERSION"  ]; then
+    cd releasenotes-builder
+    mvn versions:set-property -Dproperty=checkstyle.version -DnewVersion="$LATEST_VERSION"
+    echo "Version updated to $LATEST_VERSION"
+    git config --global user.name 'github-actions[bot]'
```
romani, line 18: please move commit commands to github action.
Shell scripts should only update target version,
commit should be  single for whole repo. 

in Github Action code you should check if changes exists before doing commit.

I donot see a reason to do MVN_CURRENT_VERSION, just do update on what is take from metadata file, if content is same "git diff" will be empty, so commit will be skipped.

romani (changes requested):
items:

.ci/bump-releasenotes-builder-version.sh
romani: script should take version as argument to let Github action take it ones from metadata file and all scripts just update version without going to internet.

.ci/bump-releasenotes-builder-version.sh
Rahulkhinchi03: 1. Commit commands moved to GitHub Action.
2. Shell script code updated to the only updated target version.
3. `MVN_CURRENT_VERSION` removed.
4. GitHub Action commit and condition to check if changes exist code added.

.ci/bump-releasenotes-builder-version.sh
Rahulkhinchi03: Changes Done

romani (approved):
ok to merge

PR #636, opened 2022-08-19
Issue #634: satelite projects should use latest checkstyle as soon at is appears in maven repo

### LINKED ISSUE #634: satelite projects should use latest checkstyle as soon at is appears in maven repo

Rahulkhinchi03:
Partially Fixes #634

### DIFF
--- a/.ci/bump-cs-version-in-patch-diff-report-tool.sh
+++ b/.ci/bump-cs-version-in-patch-diff-report-tool.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+set -e
+
+cd patch-diff-report-tool
+mvn versions:set-property -Dproperty=checkstyle.version -DnewVersion="$1"
+echo "Version updated to $1"


.ci/bump-patch-tool-version.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+set -e
```
romani, line 2: apply comments from https://github.com/checkstyle/contribution/pull/635#pullrequestreview-1079599409

romani (changes requested):
item:

.ci/bump-patch-tool-version.sh
Rahulkhinchi03: Changes Done

romani (approved):
ok to merge

Issue #614, opened 2022-07-02
Suppress all Idea Inspection violations


nick-mancuso:
Noticed at https://teamcity.jetbrains.com/viewLog.html?buildId=3876788&buildTypeId=Checkstyle_ContributionIdeaInspectionsPullRequest and in all PRs, Idea Inspections is failing from a bunch of errors and warnings. We need to suppress all to fix damaged master build.

PR #642, opened 2022-08-26
Issue #614: Resolve Deprecated API IDEA violations

Vyom-Yadav:
#614
From `org.eclipse.jgit.lib.Repository.peel(..)` javadoc:
```java
/**                                                                          
 * Peel a possibly unpeeled reference to an annotated tag.                   
 * <p>                                                                       
 * If the ref cannot be peeled (as it does not refer to an annotated tag)    
 * the peeled id stays null, but {@link org.eclipse.jgit.lib.Ref#isPeeled()} 
 * will be true.                                                             
 *                                                                           
 * @param ref                                                                
 *            The ref to peel                                                
 * @return <code>ref</code> if <code>ref.isPeeled()</code> is true; else a   
 *         new Ref object representing the same data as Ref, but isPeeled()  
 *         will be true and getPeeledObjectId will contain the peeled object 
 *         (or null).                                                        
 * @deprecated use {@code getRefDatabase().peel(ref)} instead.               
 */                                                                          
```

### DIFF
--- a/releasenotes-builder/src/main/java/com/github/checkstyle/github/NotesBuilder.java
+++ b/releasenotes-builder/src/main/java/com/github/checkstyle/github/NotesBuilder.java
@@ -274,7 +274,7 @@ public final class NotesBuilder {
             actualObjectId = repo.resolve(ref);
         }
         else {
-            final Ref repoPeeled = repo.peel(referenceObj);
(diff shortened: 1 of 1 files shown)


romani (approved):
Please rerun drone.io master build after update, to recheck release notes

nick-mancuso (approved):

Issue #652, opened 2022-09-04
releasenotes-builder: Release notes for GitHub shows escape at-symbol


Rahulkhinchi03:
https://github.com/checkstyle/checkstyle/pull/12007#issuecomment-1221431907

romani:
@stoyanK7 , can you help with this issue?

stoyanK7:
@romani I am actually working on it. Submitting a pull request today.

romani:
@stoyanK7 , please do not hesitate to send draft PR to let us see in what direction you are going.

rnveach:
Fix was merged

Issue #612, opened 2022-06-15
releasenotes-builder: Update releasenotes-builder to generate content to follow indentation in LI tag


romani:
We need to updated code to do smart wrap of long lines into few lines with proper indentation.

if that is problem due to template engine we use, so I am ok with post generation processing of result xml.

```
    <section name="Release 10.4">
      <div class="releaseDate">30.07.2022</div>
      <p>Bug fixes:</p>
        <ul>
          <li>
            MissingJavadocType: Support qualified annotation names.
            Author: stoyanK7
            <a href="https://github.com/checkstyle/checkstyle/issues/11736">#11736</a>
          </li>
          <li>
            Update google_checks.xml to have the SuppressionCommentFilter and
SuppressWarningsHolder modules in the config by default (and by
extension, SuppressWarningsFilter).
            Author: Daniel del Castillo
            <a href="https://github.com/checkstyle/checkstyle/issues/11655">#11655</a>
          </li>
        </ul>
      <p>Notes:</p>

```

expected:
```
    <section name="Release 10.3.2">
      <div class="releaseDate">31.07.2022</div>
      <p>Bug fixes:</p>
        <ul>
          <li>
            MissingJavadocType: Support qualified annotation names.
            Author: stoyanK7
            <a href="https://github.com/checkstyle/checkstyle/issues/11736">#11736</a>
          </li>
          <li>
            Update google_checks.xml to have the SuppressionCommentFilter and
            SuppressWarningsHolder modules in the config by default (and by
            extension, SuppressWarningsFilter).
            Author: Daniel del Castillo
            <a href="https://github.com/checkstyle/checkstyle/issues/11655">#11655</a>
          </li>
        </ul>

```

rnveach:
> generated content is violating limit of 100 symbols 

@romani Can you add examples? I thought we fixed most issues in https://github.com/checkstyle/contribution/issues/312 . Also not sure if this somehow applies to https://github.com/checkstyle/contribution/issues/364 or what.

romani:
I updated description to describe actual problem

Rahulkhinchi03:
@romani Do you want me to write some java code for this or we can do this after the post is created and the shell script focuses on the indentation? I believe shell script can do it more smartly.

Please share your thoughts!

romani:
Shell script unlikely make it smart.
Please try to do this in java code of releasenotes builder.

Rahulkhinchi03:
[...]

Issue #364, opened 2019-04-01
releasenotes-builder: avoid extra line wrapping


romani:
https://github.com/checkstyle/contribution/issues/312#issuecomment-478472949

It is better to not wrap line if it could fit in line limit 100 symbols per line. Xdoc file has certain indentation that unlikely to change.

Additional minor issue is to trim spaces in front of issue title, it is visible in non xdoc content.

rnveach:
Is it ok to just increase the limit by 10? or you want 17?
original 70 + addition of 10 + 13 indentation spaces gives us a max line of 93.

romani:
it is not a general tool, I am ok with anything, just to avoid manual update of release notes.
I still want to make release + notifications completely automated and not be locked to me.

Rahulkhinchi03:
@romani @rnveach Please confirm the limit of line.

romani:
Line limit is 100.
Any line that is above, should be wrapped with proper indentation.

Rahulkhinchi03:
Just this needs to be changed, right? 
https://github.com/checkstyle/contribution/blob/master/releasenotes-builder/src/main/java/com/github/checkstyle/globals/ReleaseNotesMessage.java#L32 and with this, the template's some minor indentation

romani:
yes, I do not remember why it was set to 70, please test after update, there might be some other change is required also.

rnveach:
> I do not remember why it was set to 70

https://github.com/checkstyle/contribution/issues/312#issuecomment-478472949

Rahulkhinchi03:
@romani @rnveach kindly help with what other changes will be required other than:
https://github.com/checkstyle/contribution/issues/364#issuecomment-1236368780

raghav-rama:
@Rahulkhinchi03 have you resolved the issue?

romani:
Issue is not resolved. You can reuse referenced code as base but it should be further improved. Look at referenced PR.

PR #670, opened 2022-10-18
Issue #648: releasenote-builder: should have execution mode to fail if release version (by pom) is not matching to issues labels

### LINKED ISSUE #648: releasenote-builder: should have execution mode to fail if next version (by pom) is not matching to issues labels
Correct: 
https://cloud.drone.io/checkstyle/checkstyle/12329/4/2 
in pom 10.3.3-SNAPSHOT
in releasenotes 
```
Checkstyle 10.3.3 - https://checkstyle.org/releasenotes.html#Release_10.3.3
Bug fixes:
  Translate messages to Russian
  RequireThisCheck not able to handle try-with-resource
  Sarif support is not extended onto the Ant task defined.
  InvalidJavadocPosition: false positive for the generic constructor without access modifier
Notes:
 ......
```

Incorrect: 
https://cloud.drone.io/checkstyle/checkstyle/12325/4/2
in pom 10.3.3-SNAPSHOT
in releasenotes 
```
Checkstyle 10.3.3 - https://checkstyle.org/releasenotes.html#Release_10.3.3
New:
  RequireThisCheck not able to handle try-with-resource
Bug fixes:
  Translate messages to Russian
  Sarif support is not extended onto the Ant task defined.
  InvalidJavadocPosition: false positive for the generic constructor without access modifier
Notes:
 ......
```

Rules is simple:
if pom version is 3 digits and in issue label generate releasenotes where is "new" or "breaking compatibility" - ERROR
if pom version is 2 digits and in issue label generate releasenotes where is only "bugs" or "Notes" - ERROR

We need new option `validateVersion`
https://github.com/checkstyle/contribution/tree/master/releasenotes-builder#command-line-usage
this validation should happen at the most end to let all results be printed and done.
This mode should be activated only in drone CI to fail in master branch build and notify maintainers that version type in pom and by labels is mismatching.
Maintainer will make decision to bump version in pom or change labels.

stoyanK7:
Resolves #648


releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
```diff
@@ -96,7 +96,7 @@ public void testGenerateOnlyBug() throws Exception {
             createNotes(null, null, createAllNotes(), null, null), createBaseCliOptions()
                 .setGenerateAll(true).build());
 
-        Assert.assertEquals("no errors", 0, errors.size());
+        Assert.assertEquals("one error", 1, errors.size());
```
stoyanK7, line 99: I checked both tests that I changed and we have to expect an error with this configuration. The other solution is to make `createBaseCliOptions` accept an optional parameter `releaseNumber` so we can change from default `1.0.0` to two-number one - `1.0`, for instance.

releasenotes-builder/src/main/java/com/github/checkstyle/MainProcess.java
```diff
@@ -79,7 +80,50 @@ public static List<String> runPostGenerationAndPublication(
...
+            = releaseNotes.containsKey(Constants.NEW_FEATURE_LABEL)
+            || releaseNotes.containsKey(Constants.NEW_MODULE_LABEL)
+            || releaseNotes.containsKey(Constants.BREAKING_COMPATIBILITY_LABEL);
+
+        final List<String> errors = new ArrayList<>();
+
+        if (amountOfNumbers == 2 && containsNewOrBreakingCompatabilityLabel
+            || amountOfNumbers == 3 && !containsNewOrBreakingCompatabilityLabel) {
+            errors.add("Validation of release notes failed. "
+                + "There is a mismatch between release version and issues labels.");
```
stoyanK7, line 112: Was wondering whether we should throw this generic message or explain a bit in depth what we are expecting depending on `releaseVersion`. Something like
```java
if (amountOfNumbers == 2 && containsNewOrBreakingCompatabilityLabel) {
    errors.add(String.format("Validation of release notes failed. Release version contains two numbers(%s) and new or breaking compatibility labels.", releaseVersion))
} else if (amountOfNumbers == 3 && !containsNewOrBreakingCompatabilityLabel) {
    errors.add(String.format("Validation of release notes failed. Release version contains three numbers(%s) and no new or breaking compatibility labels.", releaseVersion))
}
```

releasenotes-builder/README.md
```diff
@@ -111,6 +111,8 @@ Generated files will be at specified output location.
 
 **sfRssProperties** - (optional) path to a properties file for publication to RSS.
 
+**validateVersion** - (optional) validate pom version in issues labels.
```
stoyanK7, line 114: To be honest, I am not a big fan of `validateVersion` since we are not only validating pom version but also release notes labels. `validateVersionWithReleaseNotesLabels` would sound more appropriate. But these are just my thoughts.

stoyanK7:
Oh, and I noticed we want to `not zero exit code in case of violation`. However, in such case `validateNotes` has to be called outside of `runPostGenerationAndPublication` since any `publicationErrors` would not trigger `ERROR_EXIT_CODE`.

https://github.com/checkstyle/contribution/blob/876dc6a1e1226c56b0f5ac0818cfa1bbfa8aedf2/releasenotes-builder/src/main/java/com/github/checkstyle/Main.java#L69-L93

I would suggest doing something like this(lines with `// add this` are my suggestion):
```java
    public static void main(String... args) {
        int errorCounter;
        List<String> publicationErrors = null;
        List<String> validationErrors = null; // add this
        try {
            final CliProcessor cliProcessor = new CliProcessor(args);
            cliProcessor.process();
            if (cliProcessor.hasErrors()) {
                printListOf(cliProcessor.getErrorMessages());
                errorCounter = cliProcessor.getErrorMessages().size();
            }
            else {
                final CliOptions cliOptions = cliProcessor.getCliOptions();
                final Result notesBuilderResult = runGithubNotesBuilder(cliOptions);
                errorCounter = notesBuilderResult.getErrorMessages().size();
                if (errorCounter == 0) {
                    publicationErrors = MainProcess.runPostGenerationAndPublication(
                        notesBuilderResult.getReleaseNotes(), cliOptions);
                    validationErrors = MainProcess.validateNotes( // add this
                        notesBuilderResult.getReleaseNotes(), cliOptions); // add this
                }
            }
        }
        catch (ParseException | GitAPIException | IOException | TemplateException ex) {
            errorCounter = 1;
            ex.printStackTrace();
            CliProcessor.printUsage();
        }
        if (errorCounter == 0) {
            if (publicationErrors != null && !publicationErrors.isEmpty()) {
                System.out.printf("%nPublication ends with %d errors:%n",
                                  publicationErrors.size());
                printListOf(publicationErrors);
            }
            else {
                System.out.printf("%nExecution succeeded!%n");
            }
        }
        else if (validationErrors != null && !validationErrors.isEmpty()) { // add this
            System.out.printf("%nGeneration ends with validation errors.%n"); // add this
            System.exit(ERROR_EXIT_CODE); // add this
        } // add this
        else {
            System.out.printf("%nGeneration ends with %d errors.%n",
                              errorCounter);
            System.exit(ERROR_EXIT_CODE);
        }
    }
```

releasenotes-builder/src/main/java/com/github/checkstyle/CliProcessor.java
```diff
@@ -296,6 +300,8 @@ private static Options buildOptions() {
             "Access token secret for Twitter.");
         options.addOption(OPTION_TWITTER_PROPERTIES, true,
             "Properties for publication on Twitter.");
+        options.addOption(OPTION_VALIDATE_VERSION, false,
+            "Whether pom version should be validated in issues labels.");
         return options;
```
romani, line 305: We kind of do not care about pom,
We validate `options.addOption(OPTION_RELEASE_NUMBER, true, "Release number.")`

romani (changes requested):
Items

releasenotes-builder/src/main/java/com/github/checkstyle/MainProcess.java
romani: It is very bad practice to do logic based on exceptions.
We can create class to keep violation message, or simply keep list of strings - errors

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
```diff
@@ -93,7 +93,7 @@ public void testGenerateOnlyNewFeature() throws Exception {
     @Test
     public void testGenerateOnlyBug() throws Exception {
         final List<String> errors = MainProcess.runPostGenerationAndPublication(
-            createNotes(null, null, createAllNotes(), null, null), createBaseCliOptions()
+            createNotes(null, null, createAllNotes(), null, null), createBaseCliOptions("1.0")
```
romani, line 96: Can we keep generate All tests untouched?
Validation has nothing about generation.

romani (changes requested):
Items

romani:
>prefer doing this, but let me know what you think.

Can we simply have list of errors and print them at the end ? without checking nature of them

releasenotes-builder/src/main/java/com/github/checkstyle/CliProcessor.java
stoyanK7: Is `Whether release notes labels should be validated based on release number.` better?

releasenotes-builder/README.md
stoyanK7: Alright, for internal use makes sense.

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
stoyanK7: Almost all, except `testGenerateOnlyMisc()` and `testGenerateOnlyBug()`. As they are right now, they violate this new flag that we are introducing. They have a 3 number version and no `new` or `breaking compatibility` labels. Either we expect an error in the test or change the release version in those.

Will change it to expect one error.

releasenotes-builder/src/main/java/com/github/checkstyle/CliProcessor.java
romani: Whether release notes labels should be validated based on release number.

releasenotes-builder/src/main/java/com/github/checkstyle/MainProcess.java
stoyanK7: Changed to keep list of strings - errors. Done.

releasenotes-builder/src/main/java/com/github/checkstyle/CliProcessor.java
stoyanK7: Done.

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
stoyanK7: Done.

stoyanK7:
> Can we simply have list of errors and print them at the end ? without checking nature of them
 
 Done.

releasenotes-builder/src/main/java/com/github/checkstyle/CliProcessor.java
```diff
@@ -296,6 +300,8 @@ private static Options buildOptions() {
             "Access token secret for Twitter.");
         options.addOption(OPTION_TWITTER_PROPERTIES, true,
             "Properties for publication on Twitter.");
+        options.addOption(OPTION_VALIDATE_VERSION, false,
+            "Whether release notes labels should be validated based on release number.");
```
romani, line 304: It vise versa.
We validate release number to match what we have in notes.
Notes are based on commits, we can not change them.
All we can do and should do is correct release number.

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
```diff
@@ -111,7 +111,7 @@ public void testGenerateOnlyMisc() throws Exception {
             createNotes(null, null, null, createAllNotes(), null), createBaseCliOptions()
                 .setGenerateAll(true).build());
 
-        Assert.assertEquals("no errors", 0, errors.size());
+        Assert.assertEquals("one error", 1, errors.size());
```
romani, line 114: Can we detach validation from GenerateAll ?
I still think it is very separate mode of execution.

romani (changes requested):
Items

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
romani: So all existing tests will work as before and new tests will do .setvalidate().build())

releasenotes-builder/src/main/java/com/github/checkstyle/MainProcess.java
```diff
@@ -79,7 +80,50 @@ public static List<String> runPostGenerationAndPublication(
...
+            releaseNotes.containsKey(Constants.NEW_FEATURE_LABEL)
+            || releaseNotes.containsKey(Constants.NEW_MODULE_LABEL)
+            || releaseNotes.containsKey(Constants.BREAKING_COMPATIBILITY_LABEL);
+
+        final List<String> errors = new ArrayList<>();
+
+        if (amountOfCommas == 1 && containsNewOrBreakingCompatabilityLabel
+            || amountOfCommas == 2 && !containsNewOrBreakingCompatabilityLabel) {
+            errors.add("Validation of release notes failed. "
+                + "There is a mismatch between release version and issues labels.");
```
romani, line 112: Let's be more detailed on what is problem
To make message special for each invalid scenario.
To let maintainer not think and simply follow suggestion.

releasenotes-builder/src/main/java/com/github/checkstyle/MainProcess.java
```diff
@@ -79,7 +80,50 @@ public static List<String> runPostGenerationAndPublication(
...
+
+        final boolean containsNewOrBreakingCompatabilityLabel =
+            releaseNotes.containsKey(Constants.NEW_FEATURE_LABEL)
+            || releaseNotes.containsKey(Constants.NEW_MODULE_LABEL)
+            || releaseNotes.containsKey(Constants.BREAKING_COMPATIBILITY_LABEL);
+
+        final List<String> errors = new ArrayList<>();
+
+        if (amountOfCommas == 1 && containsNewOrBreakingCompatabilityLabel
+            || amountOfCommas == 2 && !containsNewOrBreakingCompatabilityLabel) {
```
romani, line 110: Can we make methods like isMinor isPatch to explain magic numbers by method names

romani (changes requested):
Items

releasenotes-builder/src/main/java/com/github/checkstyle/MainProcess.java
stoyanK7: Here are my suggestions for both. Let me know what you think:
- `isPatch(amountOfCommas) && containsNewOrBreakingCompatabilityLabel`
  - `Validation of release number failed. Release number is a patch(10.4), but release notes do not contain 'new' or 'breaking compatability' labels.`
- `isMinor(amountOfCommas) && !containsNewOrBreakingCompatabilityLabel`
  - `Validation of release number failed. Release number is minor(10.3.5), but release notes contain 'new' or 'breaking compatability' labels.`

releasenotes-builder/src/main/java/com/github/checkstyle/MainProcess.java
stoyanK7: Done.

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
stoyanK7: Done.

releasenotes-builder/src/main/java/com/github/checkstyle/CliProcessor.java
stoyanK7: Was a bit confused since you mention `Maintainer will make decision to bump version in pom or change labels.` in original issue. Fixed though. Done.

releasenotes-builder/src/main/java/com/github/checkstyle/MainProcess.java
```diff
@@ -79,7 +80,81 @@ public static List<String> runPostGenerationAndPublication(
             Multimap<String, ReleaseNotesMessage> releaseNotes, CliOptions cliOptions)
             throws IOException, TemplateException {
         runPostGeneration(releaseNotes, cliOptions);
-        return runPostPublication(cliOptions);
+        final List<String> errors = new ArrayList<>();
+        if (cliOptions.isValidateVersion() != null) {
```
stoyanK7, line 84: I am having trouble finding the place to set the default value of `validateVersion` to `false`. I see that some values do get a default value, others `null`.

stoyanK7: Nevermind, found it.

releasenotes-builder/src/main/java/com/github/checkstyle/MainProcess.java
romani: Yes.

`Please correct release number by running https://github.com/checkstyle/checkstyle/actions/workflows/bump-version-and-update-milestone.yml`

romani: FYI https://github.com/checkstyle/checkstyle/wiki/Semi-automated-release

releasenotes-builder/src/main/java/com/github/checkstyle/CliProcessor.java
romani: You are correct, but usually it results in correction of version number

releasenotes-builder/src/main/java/com/github/checkstyle/MainProcess.java
```diff
@@ -79,7 +80,82 @@ public static List<String> runPostGenerationAndPublication(
...
+
+    /**
+     * Returns {@code true} if amount of commas is 1. This means
+     * the release is a patch.
+     *
+     * @param amountOfCommas amount of commas in release number.
+     * @return {@code true} if amount of commas is 1.
+     */
+    private static boolean isPatch(long amountOfCommas) {
+        return amountOfCommas == 1;
```
romani, line 158: One comma is minor.
https://semver.org/
https://github.com/checkstyle/checkstyle/issues/3709

romani (changes requested):
Items

releasenotes-builder/src/main/java/com/github/checkstyle/MainProcess.java
stoyanK7: Didn't know that. Thanks. Should be fixed. Done.

releasenotes-builder/src/main/java/com/github/checkstyle/MainProcess.java
stoyanK7: Done.

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
```diff
@@ -262,6 +262,34 @@ public void testGenerateCustomTemplate() throws Exception {
...
+                + "actions/workflows/bump-version-and-update-milestone.yml");
+    }
+
+    @Test
+    public void testValidateNotesPatch() throws Exception {
+        final List<String> errors = MainProcess.runPostGenerationAndPublication(
+            createNotes(createAllNotes(), null, null, null, null), createBaseCliOptions("1.0.0")
+                .setValidateVersion(true).build());
+
+        Assert.assertEquals("one error", 1, errors.size());
```
romani, line 285: Please add tests for no violations for minor and patch

romani (changes requested):
Items

romani:
Please send PR to main repo to activate validation
https://github.com/checkstyle/checkstyle/blob/ad27aa0feb4804cf3a14430b2cb90c3d13dbe532/.drone.yml#L62

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
stoyanK7: Done.

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
```diff
@@ -262,6 +262,52 @@ public void testGenerateCustomTemplate() throws Exception {
...
+            "Validation of release number failed. Release number is a patch(1.0.0), but "
+                + "release notes contain 'new' or 'breaking compatability' labels. Please correct "
+                + "release number by running https://github.com/checkstyle/checkstyle/actions/"
+                + "workflows/bump-version-and-update-milestone.yml");
+    }
+
+    @Test
+    public void testValidateNotesPatchSuccess() throws Exception {
+        final List<String> errors = MainProcess.runPostGenerationAndPublication(
+            createNotes(null, null, null, null, null), createBaseCliOptions("1.0.0")
```
romani, line 305: Please add in addition, few test with valid existing, real notes in all goups.

romani (changes requested):
Items

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
stoyanK7: Done.

romani (approved):
Awesome work.

stoyanK7:
@romani Thanks. Which issue shall we tackle next?

romani:
https://github.com/checkstyle/contribution/issues/652
https://github.com/checkstyle/contribution/issues/364
https://github.com/checkstyle/contribution/issues/612

PR #676, opened 2022-10-28
Pull #676: Fix releasenotes-builder: print validation error messages

stoyanK7:
Prints validation errors when there are such. I do not know why it is not taken care of just like with `publicationErrors` but whatever.

---
From https://github.com/checkstyle/checkstyle/pull/12316#issuecomment-1292582626
> Though there is still a small problem. The error message is not printed: https://cloud.drone.io/checkstyle/checkstyle/13134/4/2. 

---
Output: 
```
[WARN] Issue #11214 "Specify violation messages in input files." is not closed! Please review issue https://github.com/checkstyle/checkstyle/issues/11214

[ERROR] Issue #12243 does not have breaking compatibility, bug, miscellaneous, new feature, new module label! Please set label at https://github.com/checkstyle/checkstyle/issues/12243

Validation of release number failed. Release number is a patch(10.3.5), but release notes contain 'new' or 'breaking compatability' labels. Please correct release number by running https://github.com/checkstyle/checkstyle/actions/workflows/bump-version-and-update-milestone.yml

Generation ends with 2 errors.

```


releasenotes-builder/src/main/java/com/github/checkstyle/Main.java
```diff
@@ -69,6 +69,7 @@ public static void main(String... args) {
                     final List<String> validationErrors = MainProcess
                         .validateNotes(notesBuilderResult.getReleaseNotes(), cliOptions);
                     notesBuilderResult.addErrors(validationErrors);
+                    printListOf(validationErrors);
```
rnveach, line 72: `runGithubNotesBuilder` has similar code on line 122 and 125.

Why don't we just remove that code, and move it to the end of this method so the last thing we do before ending is display all warnings and errors? This is what we seem to do with `publicationErrors`.

rnveach (changes requested):

releasenotes-builder/src/main/java/com/github/checkstyle/Main.java
romani: Yes, this becomes tricky.
We have violations that we need to print before release notes validation and printing notes. As we can not do release number validation if label is not set.

We need to print results of release notes publishing.

At the very end we need to do validation of release number and print violation on it. Violation on release number should be after print of release notes to let maintainer see what should be fixed - release number or change labels.

@stoyanK7 , can we archive above sequence?
I am ok with more refactoring or keeping two different set of errors in result object.

releasenotes-builder/src/main/java/com/github/checkstyle/Main.java
```diff
@@ -53,36 +54,37 @@ private Main() {
...
             cliProcessor.process();
             if (cliProcessor.hasErrors()) {
-                printListOf(cliProcessor.getErrorMessages());
-                errorCounter = cliProcessor.getErrorMessages().size();
+                errors.addAll(cliProcessor.getErrorMessages());
             }
             else {
                 final CliOptions cliOptions = cliProcessor.getCliOptions();
                 final Result notesBuilderResult = runGithubNotesBuilder(cliOptions);
+                errors.addAll(notesBuilderResult.getWarningMessages());
```
rnveach, line 69: Warnings shouldn't be errors, otherwise, the name should change.

stoyanK7: Would renaming to `errorsAndWarnings` be better in your opinion?

rnveach: It depends on how you plan to use it. If your using it to count errors, then no, as warnings as not errors. We shouldn't fail the utility when a warning is printed.

Why can't we keep them separate like how it was done before?

stoyanK7: You have a point, warnings have to be printed as before since we do not fail if there are such.

stoyanK7: Done. Reverted back the warnings.

stoyanK7:
Current output:
```
[WARN] Issue #11214 "Specify violation messages in input files." is not closed! Please review issue https://github.com/checkstyle/checkstyle/issues/11214

Generation ends with 2 errors.

[ERROR] Issue #12243 does not have breaking compatibility, bug, miscellaneous, new feature, new module label! Please set label at https://github.com/checkstyle/checkstyle/issues/12243
[ERROR] Validation of release number failed. Release number is a patch(10.3.5), but release notes contain 'new' or 'breaking compatability' labels. Please correct release number by running https://github.com/checkstyle/checkstyle/actions/workflows/bump-version-and-update-milestone.yml

```

releasenotes-builder/src/main/java/com/github/checkstyle/Main.java
```diff
@@ -53,36 +54,36 @@ private Main() {
...
                 if (cliOptions.isValidateVersion()) {
                     final List<String> validationErrors = MainProcess
                         .validateNotes(notesBuilderResult.getReleaseNotes(), cliOptions);
-                    notesBuilderResult.addErrors(validationErrors);
+                    errors.addAll(validationErrors);
                 }
-                errorCounter = notesBuilderResult.getErrorMessages().size();
-                if (errorCounter == 0) {
+                if (errors.isEmpty()) {
                     publicationErrors = MainProcess.runPostGenerationAndPublication(
```
romani, line 76: we should run actual generation of content to show maintainer grouped changes.
if we just print error on version it will be hard to make decision on what to do.
we can use skip publications only if error exists, so we need to split runPostGenerationAndPublication into two methods to make it sure that in case of any problem/error we can generate some files(release notes content) but we do not leak them to public.

releasenotes-builder/src/main/java/com/github/checkstyle/Main.java
```diff
@@ -94,7 +95,8 @@ public static void main(String... args) {
         }
         else {
             System.out.printf("%nGeneration ends with %d errors.%n",
-                              errorCounter);
+                              errors.size());
+            printListOf(errors);
```
romani, line 99: old logic is very complicated that confuse us a lot.

we need simply:
if exception is thrown, put content of `ex.printStackTrace();` to `errors`. `hasThrownException` will not be required.
if case `error` is not empty we print all errors and exit with error code.

if `publicationErrors` exists they are also erros, and it should fail execution also. Such old logic on `publicationErrors` might exist due to old problem to do publication and this is how we tried to suppress them, as we couldnot resolve them. Now we resolved all problems and publications must work and must fail execution if any problems happens.

romani (changes requested):
items:

releasenotes-builder/src/main/java/com/github/checkstyle/Main.java
stoyanK7: Facing some trouble adding this stack trace to the errors. Apparently [there is no easy way to convert stack trace to string](https://stackoverflow.com/questions/1149703/how-can-i-convert-a-stack-trace-to-a-string)

We either do this
```
StringWriter sw = new StringWriter();
PrintWriter pw = new PrintWriter(sw);
e.printStackTrace(pw);
String sStackTrace = sw.toString(); // stack trace as a string
```

or use [org.apache.commons.lang.exception.ExceptionUtils.getStackTrace(Throwable)](http://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/exception/ExceptionUtils.html#getStackTrace(java.lang.Throwable)).

Both of them do not pass checkstyle so we have to allow one. Which do you prefer?
```
[ERROR] [checkstyle] [ERROR] /media/stoyan/Elements/Open Source/contribution/releasenotes-builder/src/main/java/com/github/checkstyle/Main.java:27:1: Disallowed import - org.apache.commons.lang.exception.ExceptionUtils. [ImportControlMain]
```
```
[ERROR] [checkstyle] [ERROR] /media/stoyan/Elements/Open Source/contribution/releasenotes-builder/src/main/java/com/github/checkstyle/Main.java:23:1: Disallowed import - java.io.PrintWriter. [ImportControlMain]
[ERROR] [checkstyle] [ERROR] /media/stoyan/Elements/Open Source/contribution/releasenotes-builder/src/main/java/com/github/checkstyle/Main.java:24:1: Disallowed import - java.io.StringWriter. [ImportControlMain]

```

releasenotes-builder/src/main/java/com/github/checkstyle/Main.java
stoyanK7: `runPostGenerationAndPublication` consists of two methods `runPostGeneration` and `runPostPublication`. From what I see `runPostGeneration` does not return errors but rather throw exceptions so we would not even get to running `runPostPublication` if any error is thrown. Maybe just run `runPostGenerationAndPublication` without that `if (errors.isEmpty()) {` ?

releasenotes-builder/src/main/java/com/github/checkstyle/Main.java
romani: Can we print exception as is and put in error messages string like "exception happened, se details above "

releasenotes-builder/src/main/java/com/github/checkstyle/Main.java
romani: It might protect execution of this method is any error happen before.
Like if no label is found, how we can do posts generation?

releasenotes-builder/src/main/java/com/github/checkstyle/Main.java
stoyanK7: We can do that, yes.

romani:
I hope during this issue you clearly see how pure junit tests are bad in easy and actual defining of quality and behavior of application. So easy to be in illusion that all is fine.
In main projects we abandoned and forbitting writing pure unit tests and require "close to real" execution of whole module.

releasenotes-builder/src/main/java/com/github/checkstyle/Main.java
stoyanK7: I am not sure I understand what you mean exactly.

romani: Please make sure that publication doesn't happen if there is any error  happen before.

stoyanK7: This is going to require a bit of refactoring. I now see the problem with unit tests and why you try to avoid them,

releasenotes-builder/src/main/java/com/github/checkstyle/Main.java
stoyanK7: Done.

releasenotes-builder/src/main/java/com/github/checkstyle/Main.java
stoyanK7: Moved logic back to `runPostGenerationAndPublication` and added additional `boolean` parameter to tell if app has encountered any errors before. Personally, not a big fan of that and having a `boolean` parameter but it was the least amount of refactoring. Another solution would be to pass the `error` list to `runPostGenerationAndPublication` so it can check that for itself. Or maybe the best would be to get rid of `runPostGenerationAndPublication` altogether and call `runPostGeneration()` and `runPostPublication()` separately in `Main.main()`. That would require to call them separately int tests as well.

romani: If any error happen before publication, we should not publicate

romani: >That would require to call them separately int tests as well.

No problems, tests always ok to change.

stoyanK7: On it.

releasenotes-builder/src/main/java/com/github/checkstyle/MainProcess.java
```diff
@@ -72,15 +72,24 @@ private MainProcess() {
...
+            Multimap<String, ReleaseNotesMessage> releaseNotes, CliOptions cliOptions,
+            boolean hasEncounteredErrorsBefore)
             throws IOException, TemplateException {
         runPostGeneration(releaseNotes, cliOptions);
-        return runPostPublication(cliOptions);
+        final List<String> errors = new ArrayList<>();
+        if (cliOptions.isValidateVersion()) {
+            errors.addAll(validateNotes(releaseNotes, cliOptions));
+        }
+        if (!hasEncounteredErrorsBefore) {
```
romani, line 89: Rename to shouldRunPublication

romani (changes requested):
Items

releasenotes-builder/src/main/java/com/github/checkstyle/MainProcess.java
stoyanK7: Done. Also added the check for `errors.isEmpty()` we forgot.

romani (approved):
Ok to merge

releasenotes-builder/src/main/java/com/github/checkstyle/MainProcess.java
```diff
@@ -100,7 +109,7 @@ public static List<String> validateNotes(Multimap<String, ReleaseNotesMessage> r
                 || releaseNotes.containsKey(Constants.BREAKING_COMPATIBILITY_LABEL);
 
         final List<String> errors = new ArrayList<>();
-        final String errorBeginning = "Validation of release number failed.";
+        final String errorBeginning = "[ERROR] Validation of release number failed.";
```
rnveach, line 112: I'm slightly confused by this change and a similar else where. Do we not automatically say each item is an error when we print it and this is an example where we forgot? If so, sounds to me like the error printing should be doing this for us and we shouldn't be doing this manually.

I am ok if we want to delay this.

releasenotes-builder/src/main/java/com/github/checkstyle/Main.java
```diff
@@ -121,9 +106,6 @@ private static Result runGithubNotesBuilder(CliOptions cliOptions)
         if (result.hasWarnings()) {
             printListOf(result.getWarningMessages());
         }
-        if (result.hasErrors()) {
```
rnveach, line 124: We removed errors here to avoid double printing them. We may just want to consider for a future change to do the same with warnings, and then anything we call can print errors or warnings and we don't have to deal with individual lists.

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
```diff
@@ -52,12 +52,12 @@
 public class TemplateProcessorTest {
```
rnveach, line 52: I am not seeing any new tests especially using `false` for the new boolean parameter in `runPostGenerationAndPublication`. Is it possible to add some tests?

releasenotes-builder/src/main/java/com/github/checkstyle/MainProcess.java
romani: I do not worry about this, I hope we will not update this project after this iteration.
It just need to do a job.

releasenotes-builder/src/main/java/com/github/checkstyle/Main.java
romani: Better to deal with it later on  as problem comeup

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
romani: I am not sure we have any test on publication as we need to mock Twitter API, it doesn't worth to do this, too expensive for now.

rnveach (approved):

Issue #448, opened 2020-02-12
Resolve codenarc violations


romani:
after failure https://travis-ci.org/checkstyle/contribution/jobs/648904488#L222
codenarc was disabled and we leaked  following violations so we need to resolve them

```
CodeNarc Report - Nov 11, 2021 10:40:36 AM

Summary: TotalFiles=1 FilesWithViolations=1 P1=0 P2=13 P3=17

diff.groovy
    Violation: Rule=UnusedVariable P=2 Line=88 Msg=[The variable [extraMvnRegressionOptions] in class None is not used] Src=[def extraMvnRegressionOptions = cliOptions.extraMvnRegressionOptions]
    Violation: Rule=Instanceof P=2 Line=172 Msg=[The instanceof operator is used in class None] Src=[configFilesList.removeIf { it instanceof Boolean };]
    Violation: Rule=UnusedVariable P=2 Line=340 Msg=[The variable [cloneCmd] in class None is not used] Src=[def cloneCmd = '']
    Violation: Rule=ClassForName P=2 Line=455 Msg=[Violation in class None. Methods calls to Class.forName(...) can create resource leaks and should almost always be replaced with calls to ClassLoader.loadClass(...)] Src=[def textTransform = Class.forName("com.github.checkstyle.site.TextTransform").newInstance()]
    Violation: Rule=DuplicateNumberLiteral P=2 Line=482 Msg=[Duplicate Number Literal: 2] Src=[summaryIndexHtml << (" ( &#177;${diffCount[0]}, <span style=\"color: red;\">-${diffCount[2]}</span> )")]
    Violation: Rule=LineLength P=2 Line=482 Msg=[The line exceeds 120 characters. The line is 123 characters.] Src=[                    summaryIndexHtml << (" ( &#177;${diffCount[0]}, <span style=\"color: red;\">-${diffCount[2]}</span> )")]
    Violation: Rule=DuplicateNumberLiteral P=2 Line=484 Msg=[Duplicate Number Literal: 2] Src=[else if (diffCount[2] == 0) {]
    Violation: Rule=DuplicateNumberLiteral P=2 Line=488 Msg=[Duplicate Number Literal: 2] Src=[summaryIndexHtml << (" ( &#177;${diffCount[0]}, <span style=\"color: red;\">-${diffCount[2]}</span>, <span style=\"color: green;\">+${diffCount[1]}</span> )")]
[...]

PR #695, opened 2022-11-11
Issue #448: resolve more codenarcs

rnveach:
Issue #448

### DIFF
--- a/checkstyle-tester/diff.groovy
+++ b/checkstyle-tester/diff.groovy
@@ -453,7 +453,7 @@ def getTextTransform() {
         .toFile()
     def diffToolJarPath = getPathToDiffToolJar(diffToolDir)
     this.class.classLoader.rootLoader.addURL(new URL("file:$diffToolJarPath"))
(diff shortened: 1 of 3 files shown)


checkstyle-tester/diff.groovy
```diff
@@ -861,3 +861,7 @@ class CheckstyleReportInfo {
         this.commitTime = commitTime
     }
 }
+
+class Constants {
+    static final TWO = 2
```
romani, line 866: Do we really need to make our code terrible?
I would rather keep suppression or disable rule

romani (changes requested):
Items

rnveach:
https://app.travis-ci.com/github/checkstyle/contribution/jobs/588259391
Retrying.

rnveach:
Failure is stable but no reason why is given. Only change is the class loading now.

romani (approved):

rnveach:
It is passing now.

PR #702, opened 2022-11-23
Pull #702: Fail releasenotes generator if double quotes exist in release notes

stoyanK7:
From https://github.com/checkstyle/checkstyle/pull/12407#discussion_r1027298821

> We need to update relasenotes generator to fail with error if such symbol exists or escape it as we did with @.
We can not escape " in release notes builder, as content of notes should be copy-paste able by user to github.
we need to escape only because we send content by json format (nuances of sending data, not a content of itself).


stoyanK7:
@romani I am unsure of the path I am taking here. Is this okay?

romani:
Hmmm, why escaping by jq does not work?

Content can be any, Json can handle all if we escape double quote.
I am missing something?

PR #703, opened 2022-11-24
minor: correct release notes options

rnveach:
Missing options and mis-spellings. Also corrected periods.


rnveach:
I also recommend if we aren't using all these options to think about start trimming them back.

romani (approved):

PR #706, opened 2022-11-25
Issue #701: releasenotes builder: release version validation should work on digits

stoyanK7:
Resolves #701

---
For the manual test I modified the script to be pointing to this branch and changed `CS_RELEASE_VERSION` to `10.5.0`. Runs with no errors: https://github.com/stoyanK7/checkstyle/commit/dee3811dd6eb71765ccf3bc085763cfdf876042b

If I change `CS_RELEASE_VERSION` to `10.5.1`, it is failing as expected.
```
[WARN] Issue #11163 "Enforce file size on Java inputs" is not closed! Please review issue https://github.com/checkstyle/checkstyle/issues/11163
[WARN] Issue #11446 "Update Tests to use new 'verifyXxxxxx' method or 'execute' that use inlined config in Input files" is not closed! Please review issue https://github.com/checkstyle/checkstyle/issues/11446
[WARN] Issue #11214 "Specify violation messages in input files." is not closed! Please review issue https://github.com/checkstyle/checkstyle/issues/11214
[WARN] Issue #12330 "Resolve Pitest for Block Profile-2" is not closed! Please review issue https://github.com/checkstyle/checkstyle/issues/12330

Generation ends with 1 errors.

[ERROR] Validation of release number failed. Release number is a patch(10.5.1), but release notes contain 'new' or 'breaking compatability' labels. Please correct release number by running https://github.com/checkstyle/checkstyle/actions/workflows/bump-version-and-update-milestone.yml

```

PR #710, opened 2022-11-27
config: ignore .attach_pid files

stoyanK7:
Resolves https://github.com/checkstyle/contribution/pull/706#issuecomment-1328058381

> we can add it to https://github.com/checkstyle/contribution/blob/master/.gitignore to not bother us next time


romani (approved):
https://www.w3schools.com/git/git_ignore.asp?remote=github

PR #705, opened 2022-11-25
Release notes mock test

rnveach:
This is what brought the PR https://github.com/checkstyle/contribution/pull/703 . Alternate idea for https://github.com/checkstyle/contribution/issues/327#issuecomment-1326900135 .

This PR is not expected to pass CI yet, but it is a working example. This draft PR is looking for approval as being the future method for testing in this project.

Including @stoyanK7 .

This is rough example executing Main in a test with successful completion. No commits are returned, no errors produced. We can take this and make common test methods where we return specific commits, and can expand on other things like publishing failures (by throwing exceptions), and eventually test all aspects of our code.

Any API calls like Git, GitHub, Twitter, etc... are separated from the main logic and mocked to not really execute, but return results for our logic processing. They basically will not execute and we skip over them completely since they are isolated into their own world.

~Besides more clean up is needed, I believe more of GitHub needs to be mocked like issue labels and~ we should be capturing output.

PR #712, opened 2022-12-01
Issue #327: releasenotes-builder: add better in-depth tests

### LINKED ISSUE #327: releasenotes-builder: add better in-depth tests

stoyanK7:
Resolves #327

### DIFF
--- a/releasenotes-builder/src/test/java/com/github/checkstyle/MainTest.java
+++ b/releasenotes-builder/src/test/java/com/github/checkstyle/MainTest.java
@@ -19,102 +19,315 @@
 
 package com.github.checkstyle;
 
-import static org.kohsuke.github.GHIssueState.OPEN;
+import static org.kohsuke.github.GHIssueState.CLOSED;
 
+import org.junit.Assert;
 import org.junit.Test;
 
 import com.github.checkstyle.internal.AbstractReleaseNotesTestSupport;
 
 public class MainTest extends AbstractReleaseNotesTestSupport {
-
-    @Test
-    public void testValidateVersion() {
-        runMainAndAssertReturnCode(0,
-                "-releaseNumber", "10.0.1",
-                "-validateVersion"
-        );
-    }
-
     @Test
     public void testNoCommits() {
-        runMainAndAssertReturnCode(0,
-                "-localRepoPath", getTempFolder().getAbsolutePath(),
-                "-remoteRepoPath", "checkstyle/checkstyle",
-                "-startRef", "12345678",
-                "-releaseNumber", "10.0.1",
-                "-outputLocation", getTempFolder().getAbsolutePath(),
-                "-githubAuthToken", "TOKEN",
-                "-generateAll",
-                "-publishTwit",
-                "-twitterConsumerKey", "KEY",
-                "-twitterConsumerSecret", "SECRET",
-                "-twitterAccessToken", "TOKEN",
-                "-twitterAccessTokenSecret", "SECRET",
-                "-validateVersion"
+        runMainContentGenerationAndAssertReturnCode(0,
+            "-releaseNumber", "10.0.1",
+            "-generateAll",
+            "-validateVersion"
         );
+
+        Assert.assertEquals("expected error output", "", systemErr.getLog());
+        Assert.assertEquals("expected output", MSG_EXECUTION_SUCCEEDED, systemOut.getLog());
     }
 
     @Test
     public void testUnknownCommit() {
         addCommit("Hello World", "CheckstyleUser");
 
-        runMainAndAssertReturnCode(0,
-                "-localRepoPath", getTempFolder().getAbsolutePath(),
-                "-remoteRepoPath", "checkstyle/checkstyle",
-                "-startRef", "12345678",
-                "-releaseNumber", "10.0.1",
-                "-outputLocation", getTempFolder().getAbsolutePath(),
-                "-githubAuthToken", "TOKEN",
-                "-generateAll",
-                "-publishTwit",
-                "-twitterConsumerKey", "KEY",
-                "-twitterConsumerSecret", "SECRET",
-                "-twitterAccessToken", "TOKEN",
-                "-twitterAccessTokenSecret", "SECRET",
-                "-validateVersion"
+        runMainContentGenerationAndAssertReturnCode(0,
+            "-releaseNumber", "10.0.1.",
+            "-generateAll",
+            "-validateVersion"
         );
+
+        Assert.assertEquals("expected error output", "", systemErr.getLog());
+        Assert.assertEquals("expected output", MSG_EXECUTION_SUCCEEDED, systemOut.getLog());
     }
 
     @Test
     public void testIssueCommitWithIssueNotFound() {
         addCommit("Issue #1: Hello World", "CheckstyleUser");
 
-        runMainAndAssertReturnCode(-2,
-                "-localRepoPath", getTempFolder().getAbsolutePath(),
-                "-remoteRepoPath", "checkstyle/checkstyle",
-                "-startRef", "12345678",
-                "-releaseNumber", "10.0.1",
-                "-outputLocation", getTempFolder().getAbsolutePath(),
-                "-githubAuthToken", "TOKEN",
-                "-generateAll",
-                "-publishTwit",
-                "-twitterConsumerKey", "KEY",
-                "-twitterConsumerSecret", "SECRET",
-                "-twitterAccessToken", "TOKEN",
-                "-twitterAccessTokenSecret", "SECRET",
-                "-validateVersion"
+        runMainContentGenerationAndAssertReturnCode(-2,
+            "-releaseNumber", "10.0.1",
+            "-generateAll",
+            "-validateVersion"
         );
     }
 
     @Test
     public void testIssueCommit() {
         addCommit("Issue #1: Hello World", "CheckstyleUser");
-        addIssue(1, OPEN, "Hello World", MISC);
+        addIssue(1, CLOSED, "Hello World", MISC);
 
-        runMainAndAssertReturnCode(0,
-                "-localRepoPath", getTempFolder().getAbsolutePath(),
-                "-remoteRepoPath", "checkstyle/checkstyle",
-                "-startRef", "12345678",
-                "-releaseNumber", "10.0.1",
-                "-outputLocation", getTempFolder().getAbsolutePath(),
-                "-githubAuthToken", "TOKEN",
-                "-generateAll",
-                "-publishTwit",
-                "-twitterConsumerKey", "KEY",
-                "-twitterConsumerSecret", "SECRET",
-                "-twitterAccessToken", "TOKEN",
-                "-twitterAccessTokenSecret", "SECRET",
-                "-validateVersion"
+        runMainContentGenerationAndAssertReturnCode(0,
+            "-releaseNumber", "10.0.1",
+            "-generateAll",
+            "-validateVersion"
         );
+
+        Assert.assertEquals("expected error output", "", systemErr.getLog());
+        Assert.assertEquals("expected output", MSG_EXECUTION_SUCCEEDED, systemOut.getLog());
+    }
+
+    @Test
+    public void testValidateVersionMinorNoIssues() {
+        runMainContentGenerationAndAssertReturnCode(-2,
+            "-releaseNumber", "10.0.0",
+            "-generateAll",
+            "-validateVersion"
+        );
+
+        Assert.assertEquals("expected error output", "", systemErr.getLog());
+        Assert.assertEquals("expected output",
+            getExecutionFailedMessage(1) + getReleaseIsMinorMessage("10.0.0"), systemOut.getLog());
+    }
+
+    @Test
+    public void testValidateVersionMinorBreaking() {
+        addCommit("Issue #1: Hello World", "CheckstyleUser");
+        addIssue(1, CLOSED, "Hello World", BREAKING);
+
+        runMainContentGenerationAndAssertReturnCode(0,
+            "-releaseNumber", "10.0.0",
+            "-generateAll",
+            "-validateVersion"
+        );
+
+        Assert.assertEquals("expected error output", "", systemErr.getLog());
+        Assert.assertEquals("expected output", MSG_EXECUTION_SUCCEEDED, systemOut.getLog());
+    }
+
+    @Test
+    public void testValidateVersionMinorNewFeature() {
+        addCommit("Issue #1: Hello World", "CheckstyleUser");
+        addIssue(1, CLOSED, "Hello World", NEW_FEATURE);
+
+        runMainContentGenerationAndAssertReturnCode(0,
+            "-releaseNumber", "10.0.0",
+            "-generateAll",
+            "-validateVersion"
+        );
+
+        Assert.assertEquals("expected error output", "", systemErr.getLog());
+        Assert.assertEquals("expected output", MSG_EXECUTION_SUCCEEDED, systemOut.getLog());
+    }
+
+    @Test
+    public void testValidateVersionMinorBug() {
+        addCommit("Issue #1: Hello World", "CheckstyleUser");
+        addIssue(1, CLOSED, "Hello World", BUG);
+
+        runMainContentGenerationAndAssertReturnCode(-2,
+            "-releaseNumber", "10.0.0",
+            "-generateAll",
+            "-validateVersion"
+        );
+
+        Assert.assertEquals("expected error output", "", systemErr.getLog());
+        Assert.assertEquals("expected output",
+            getExecutionFailedMessage(1) + getReleaseIsMinorMessage("10.0.0"), systemOut.getLog());
+    }
+
+    @Test
+    public void testValidateVersionMinorMiscellaneous() {
+        addCommit("Issue #1: Hello World", "CheckstyleUser");
+        addIssue(1, CLOSED, "Hello World", MISC);
+
+        runMainContentGenerationAndAssertReturnCode(-2,
+            "-releaseNumber", "10.0.0",
+            "-generateAll",
+            "-validateVersion"
+        );
+
+        Assert.assertEquals("expected error output", "", systemErr.getLog());
+        Assert.assertEquals("expected output",
+            getExecutionFailedMessage(1) + getReleaseIsMinorMessage("10.0.0"), systemOut.getLog());
+    }
+
+    @Test
+    public void testValidateVersionMinorNewModule() {
+        addCommit("Issue #1: Hello World", "CheckstyleUser");
+        addIssue(1, CLOSED, "Hello World", NEW_MODULE);
+
+        runMainContentGenerationAndAssertReturnCode(0,
+            "-releaseNumber", "10.0.0",
+            "-generateAll",
+            "-validateVersion"
+        );
+
+        Assert.assertEquals("expected error output", "", systemErr.getLog());
+        Assert.assertEquals("expected output", MSG_EXECUTION_SUCCEEDED, systemOut.getLog());
+    }
+
+    @Test
+    public void testValidateVersionMinorBugAndMiscellaneous() {
+        addCommit("Issue #1: Hello World", "CheckstyleUser");
+        addCommit("Issue #2: Hello World", "CheckstyleUser");
+        addIssue(1, CLOSED, "Hello World", BUG);
+        addIssue(2, CLOSED, "Hello World", MISC);
+
+        runMainContentGenerationAndAssertReturnCode(-2,
+            "-releaseNumber", "10.0.0",
+            "-generateAll",
+            "-validateVersion"
+        );
+
+        Assert.assertEquals("expected error output", "", systemErr.getLog());
+        Assert.assertEquals("expected output",
+            getExecutionFailedMessage(1) + getReleaseIsMinorMessage("10.0.0"), systemOut.getLog());
+    }
+
+    @Test
+    public void testValidateVersionMinorNewModuleNewFeatureAndBreaking() {
+        addCommit("Issue #1: Hello World", "CheckstyleUser");
+        addCommit("Issue #2: Hello World", "CheckstyleUser");
+        addCommit("Issue #3: Hello World", "CheckstyleUser");
+        addIssue(1, CLOSED, "Hello World", NEW_MODULE);
+        addIssue(2, CLOSED, "Hello World", NEW_FEATURE);
+        addIssue(3, CLOSED, "Hello World", BREAKING);
+
+        runMainContentGenerationAndAssertReturnCode(0,
+            "-releaseNumber", "10.0.0",
+            "-generateAll",
+            "-validateVersion"
+        );
+
+        Assert.assertEquals("expected error output", "", systemErr.getLog());
+        Assert.assertEquals("expected output", MSG_EXECUTION_SUCCEEDED, systemOut.getLog());
+    }
+
+    @Test
+    public void testValidateVersionPatchNoIssues() {
+        runMainContentGenerationAndAssertReturnCode(0,
+            "-releaseNumber", "10.0.1",
+            "-generateAll",
+            "-validateVersion"
+        );
+
+        Assert.assertEquals("expected error output", "", systemErr.getLog());
+        Assert.assertEquals("expected output", MSG_EXECUTION_SUCCEEDED, systemOut.getLog());
+    }
+
+    @Test
+    public void testValidateVersionPatchBreaking() {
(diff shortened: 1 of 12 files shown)


releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
```diff
@@ -19,65 +19,136 @@
...
-    private static final String MSG_RELEASE_IS_PATCH =
-        "[ERROR] Validation of release number failed. Release number is a patch(1.0.1), but "
-            + "release notes contain 'new' or 'breaking compatability' labels. Please correct "
-            + "release number by running https://github.com/checkstyle/checkstyle/actions/"
-            + "workflows/bump-version-and-update-milestone.yml";
+    private static final MockedStatic<CsGitHub> MOCK_CS_GIT_HUB_STATIC = mockStatic(CsGitHub.class);
+
+    private static final MockedStatic<CsGit> MOCK_CS_GIT_STATIC = mockStatic(CsGit.class);
+
+    private static final MockedStatic<TwitterPublisher>
```
stoyanK7, line 100: These `MockedStatic`s need to be a class field so we can close them after all tests have been run. If not, we get `static mocking is already registered in the current thread To create a new mock, the existing static mock registration must be deregistered` error when tests in `MainTest` start to run.

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
```diff
@@ -19,65 +19,136 @@
...
         temporaryFolder.create();
+        TEST_COMMITS.clear();
+        TEST_ISSUES.clear();
+    }
+
+    @AfterClass
+    public static void close() {
+        MOCK_CS_GIT_HUB_STATIC.close();
+        MOCK_CS_GIT_STATIC.close();
+        MOCK_TWITTER_PUBLISHER_STATIC.close();
```
stoyanK7, line 134: We close the static mocks down here.

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
```diff
@@ -19,65 +19,136 @@
...
-import com.google.common.collect.ArrayListMultimap;
-import com.google.common.collect.Multimap;
+import com.github.checkstyle.git.CsGit;
+import com.github.checkstyle.github.CsGitHub;
+import com.github.checkstyle.internal.GhIssueUtil;
+import com.github.checkstyle.internal.RevCommitUtil;
+import com.github.checkstyle.publishers.TwitterPublisher;
 
 public class TemplateProcessorTest {
+    private static final Set<RevCommit> TEST_COMMITS = new LinkedHashSet<>();
```
stoyanK7, line 77: Had to use a `LinkedHashSet` instead of just `HashSet` since order of commits(and output in turn) was randomized on every run making it impossible to assert.

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
```diff
@@ -224,236 +391,315 @@ public void testGenerateCustomTemplate() throws Exception {
...
+            "-releaseNumber", "1.0.1",
+            "-outputLocation", getOutputLocation(),
+            "-githubAuthToken", "TOKEN",
+            "-generateGitHub",
+            "-validateVersion"
+        );
     }
 
     @Test
     public void testValidateNotesPatchBugAndMiscellaneous() throws Exception {
```
stoyanK7, line 646: Not going to lie, not a big fan of those `...ValidateNotes...` tests being here in `TemplateProcessorTest`. Nothing to do with testing the `TemplateProcessor`. I would personally put them in their own class like `ValidateNotesTest` or at least in `MainTest`. Let me know what you think.

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
stoyanK7: Seems like we got this error in CI as well: https://ci.appveyor.com/project/Checkstyle/contribution/builds/45550414/job/noa9wcc5lknd0b7l#L221
```
For com.github.checkstyle.github.CsGitHub, static mocking is already registered in the current thread
To create a new mock, the existing static mock registration must be deregistered
	at com.github.checkstyle.templates.TemplateProcessorTest.<clinit>(TemplateProcessorTest.java:96)
	... 24 more
```
I think this deregistering has to be done in every test class that uses static mocks.

rnveach:
The code you copied from my PR was not really meant to be duplicated verbatim in other tests, which is why, as you identified, you are seeing these failures. I was expecting it only to be used within that 1 test and anything else will require another set up instead of duplicating a bunch of the same code to every new test class.

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
romani: commits is List - just a sequence of commits of strict order.
it is not a set.

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
romani: TemplateProcessorTest should "validate/test" only output file creation.

we have validation in MainProcess.validateNotes method.
so test file should be MainProcessTest

but execution of test happens from Main class, but it is detail of implementation, some execution might be directly by MainProcess.validateNotes.

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
romani: I do not like static. Sharing fields between test methods is always a bad, and problematic in long run.

please look at my extra commit, to avoid static.

please try to convert fields to local variables, the less fields the better.
test should be something like:
```
public void testGenerateOnlyBreakingCompatibility() throws Exception {
commits=....
issues=...
setup(commits, issues)
runMainAndAssertReturnCode(
```

romani: unfortunately this is not readable test:
```
        TEST_ISSUES.addAll(createAllIssues(BUG_LABEL));
        TEST_ISSUES.addAll(createAllIssues(MISCELLANEOUS_LABEL));
        TEST_COMMITS.addAll(createAllCommits());

```

romani:
problem is with static mocks
```
testValidateNotesMinorMiscellaneous(com.github.checkstyle.templates.TemplateProcessorTest)  Time elapsed: 0.156 sec  <<< ERROR!
org.mockito.exceptions.base.MockitoException: 
For com.github.checkstyle.github.CsGitHub, static mocking is already registered in the current thread
To create a new mock, the existing static mock registration must be deregistered
	at com.github.checkstyle.templates.TemplateProcessorTest.<init>(TemplateProcessorTest.java:97)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)

```

@rnveach , how to avoid static mocks ?

rnveach:
@romani See https://github.com/checkstyle/contribution/pull/712#issuecomment-1335160430 .

Please clarify how you want tests set up. Is what I created only for `MainTest` or do we want this to be the `AbstractModuleTestSupport#verify` of release notes where every test class will use it?

romani:
>where every test class will use it?

Yes, some reusage. Test method of each target class should be in form:
Given that we have such commits and issues. When we do execution with such parameters. Then we see such results.

Here is my imaginary code for test method implementation https://github.com/checkstyle/contribution/pull/712#discussion_r1038797286

>Also please clarify how you want issues/commits to be added? Is the following ok (no more field / Util)?

Yes. But it is ok to use util methods if that makes code less verbose.

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
rnveach: @romani 
The reason this is a set is because it is what is directly returned from the production code at
https://github.com/checkstyle/contribution/blob/a0e6b5abf8c9466e5eae021cd1d6fd7902b834a4/releasenotes-builder/src/main/java/com/github/checkstyle/git/CsGit.java#L57

Even that method returns a LinkedHashSet so another issue would have to be created for it to be turned to a list if that is what is wanted.

stoyanK7:
The following test will pass if we expect `0`, which it shouldn't. Passes with `-2` as it should though. Any other expected status code fails.
```java
@Test
public void testValidateVersionMinorBug() {
    addCommit("Issue #1: Hello World", "CheckstyleUser");
    addIssue(1, OPEN, "Hello World", BUG);

    runMainAndAssertReturnCode(0,
            "-localRepoPath", getTempFolder().getAbsolutePath(),
            "-remoteRepoPath", "checkstyle/checkstyle",
            "-startRef", "12345678",
            "-releaseNumber", "10.0.0",
            "-outputLocation", getTempFolder().getAbsolutePath(),
            "-githubAuthToken", "TOKEN",
            "-generateAll",
            "-validateVersion"
    );
}
```

rnveach:
> The following test will pass if we expect 0, which it shouldn't. Passes with -2 as it should though
> but shouldn't runMainAndAssertReturnCode fail the test or am I getting it wrong?

I see the issue. We don't verify an actual exit code. I will look into a fix.

rnveach:
It seems there is a bug but I am trying to understand it more.

So far, 0 return codes can't be anything else than 0. Non-zero return code can be their non-zero or 0.

stoyanK7:
@rnveach Pushed what I have until now so you can use it. Not ready for review.

What I noticed is that if you change the expected exit code in `testGenerateOnlyBreakingCompatibility` in `TemplateProcessorTest`, `runMainAndAssertReturnCode` will fail.

rnveach:
@stoyanK7 Try this patch, this was the only way I could get it to work.

diff --git a/releasenotes-builder/src/main/java/com/github/checkstyle/Main.java b/releasenotes-builder/src/main/java/com/github/checkstyle/Main.java
index b007373..f78ef6b 100644
--- a/releasenotes-builder/src/main/java/com/github/checkstyle/Main.java
+++ b/releasenotes-builder/src/main/java/com/github/checkstyle/Main.java
@@ -76,6 +76,7 @@
         }
         if (errors.isEmpty()) {
             System.out.printf("%nExecution succeeded!%n");
+            System.exit(0);
         }
         else {
             System.out.printf("%nGeneration ends with %d errors.%n", errors.size());
diff --git a/releasenotes-builder/src/test/java/com/github/checkstyle/MainTest.java b/releasenotes-builder/src/test/java/com/github/checkstyle/MainTest.java
index 3c541f3..cc67890 100644
--- a/releasenotes-builder/src/test/java/com/github/checkstyle/MainTest.java
+++ b/releasenotes-builder/src/test/java/com/github/checkstyle/MainTest.java
@@ -29,7 +29,7 @@
 
     @Test
     public void testValidateVersion() {
-        runMainAndAssertReturnCode(0,
+        runMainAndAssertReturnCode(-2,
                 "-releaseNumber", "10.0.1",
                 "-validateVersion"
         );
@@ -117,4 +117,21 @@
                 "-validateVersion"
         );
     }
+
+    @Test
+    public void testValidateVersionMinorBug() {
+        addCommit("Issue #1: Hello World", "CheckstyleUser");
+        addIssue(1, OPEN, "Hello World", BUG);
+
+        runMainAndAssertReturnCode(-2,
+                "-localRepoPath", getTempFolder().getAbsolutePath(),
+                "-remoteRepoPath", "checkstyle/checkstyle",
+                "-startRef", "12345678",
+                "-releaseNumber", "10.0.0",
+                "-outputLocation", getTempFolder().getAbsolutePath(),
+                "-githubAuthToken", "TOKEN",
+                "-generateAll",
+                "-validateVersion"
+        );
+    }
 }
diff --git a/releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractReleaseNotesTestSupport.java b/releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractReleaseNotesTestSupport.java
index 9a88aef..99c8fe7 100644
--- a/releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractReleaseNotesTestSupport.java
+++ b/releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractReleaseNotesTestSupport.java
@@ -24,7 +24,6 @@
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.mockStatic;
-import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
@@ -177,10 +176,7 @@
         catch (Exception exception) {
             fail(String.format("Unexpected exception: %s", exception));
         }
-        if (expectedExitCode == 0) {
-            verify(mock, never()).exit(expectedExitCode);
-        }
-        else {
+        finally {
             verify(mock).exit(expectedExitCode);
         }
     }

rnveach:
@romani 
https://github.com/checkstyle/contribution/pull/705#discussion_r1034217933
> That set only version and validateVersion, all other parameters are noice in such test cases

I didn't realize it until this fix, which shows the issue I discussed at https://github.com/checkstyle/contribution/pull/712#issuecomment-1341328500 , but the other parameters are required and are not noise.

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
```diff
@@ -85,11 +82,22 @@ public void testGenerateOnlyNewFeature() throws Exception {
...
+        runMainAndAssertReturnCode(0,
+            "-localRepoPath", getTempFolder().getAbsolutePath(),
+            "-remoteRepoPath", "checkstyle/checkstyle",
+            "-startRef", "12345678",
+            "-releaseNumber", "1.0.1",
+            "-outputLocation", getTempFolder().getAbsolutePath(),
+            "-githubAuthToken", "TOKEN",
+            "-generateAll",
+            "-validateVersion"
+        );
```
romani, line 97: same big code again and again.
Should be something like:
```
runMainContentGenerationAndAssertReturnCode(0,
            "-releaseNumber", "1.0.1",
            "-generateAll",
            "-validateVersion"
        );
```
to let reader/human to focus on what matters.
if some test need more customization, it can use runMainAndAssertReturnCode.

romani (changes requested):
items:

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
stoyanK7: Done.

releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractReleaseNotesTestSupport.java
```diff
@@ -177,12 +178,34 @@ protected void runMainAndAssertReturnCode(int expectedExitCode, String... argume
...
+            Arrays.asList(
+                "-localRepoPath", getTempFolder().getAbsolutePath(),
+                "-remoteRepoPath", "checkstyle/checkstyle",
+                "-startRef", "12345678",
+                "-outputLocation", getTempFolder().getAbsolutePath(),
+                "-githubAuthToken", "TOKEN"
+            )
+        );
+        String[] arrOfArgs = new String[listOfArgs.size()];
+        arrOfArgs = listOfArgs.toArray(arrOfArgs);
```
stoyanK7, line 206: Not a big fan of what I did here, but I see no other way of appending elements to `arguments`. Open to suggestions if you see a better way. 

Shame that Java does not have a spread operator like JavaScript...

rnveach: You could do:
````
    String[] result = Arrays.copyOf(array1, array1.length + array2.length);
    System.arraycopy(array2, 0, result, array1.length, array2.length);
    return result;
````
or
````
    String[] result = new String[len1 + len2];
    System.arraycopy(array1, 0, result, 0, len1);
    System.arraycopy(array2, 0, result, len1, len2);
````

Comment (nrmancuso) [reply releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractReleaseNotesTestSupport.java:206]: you could also do something like:
```
        final List  a = List.of(...); // your args
        a.addAll(List.of(arguments)); // args passed in
        runMainAndAssertReturnCode(expectedExitCode, a.toArray(new String[0]));

```

**Review by romani — APPROVED** [review]
I am fine with update. Big leap forward from what we had.

File: releasenotes-builder/src/test/java/com/github/checkstyle/MainTest.java [hunk 7]
```diff
@@ -19,102 +19,307 @@
...
-                "-githubAuthToken", "TOKEN",
-                "-generateAll",
-                "-publishTwit",
-                "-twitterConsumerKey", "KEY",
-                "-twitterConsumerSecret", "SECRET",
-                "-twitterAccessToken", "TOKEN",
-                "-twitterAccessTokenSecret", "SECRET",
-                "-validateVersion"
+        runMainContentGenerationAndAssertReturnCode(0,
+            "-releaseNumber", "10.0.1.",
```
rnveach, line 47: How come this isn't an error? (extra period on end)

releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractPathTestSupport.java
```diff
@@ -42,6 +42,23 @@ public abstract class AbstractPathTestSupport {
     protected static final String MISC = Constants.MISCELLANEOUS_LABEL;
     protected static final String BREAKING = Constants.BREAKING_COMPATIBILITY_LABEL;
 
+    protected static final String MSG_EXECUTION_SUCCEEDED = System.lineSeparator()
+            + "Execution succeeded!" + System.lineSeparator();
+
+    protected static final String MSG_RELEASE_IS_MINOR =
+        "\nGeneration ends with 1 errors.\n\n[ERROR] Validation of release number failed. "
```
rnveach, line 49: This is not just release is minor. It has the number of errors too, which seems a bit too specific for general usage. These 2 should be split, the number of errors should either be hard coded or find some way to make it clear the number and easy to use for others.

releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractPathTestSupport.java
```diff
@@ -42,6 +42,23 @@ public abstract class AbstractPathTestSupport {
...
+            + "Release number is minor(10.0.0), but release notes do not contain 'new' or "
```
rnveach, line 50: Same with release version.

releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractPathTestSupport.java
```diff
@@ -42,6 +42,23 @@ public abstract class AbstractPathTestSupport {
...
+            + "'breaking compatability' labels. Please correct release number by running "
+            + "https://github.com/checkstyle/checkstyle/actions/workflows/bump-version-and-"
+            + "update-milestone.yml\n";
+
+    protected static final String MSG_RELEASE_IS_PATCH =
+        "\nGeneration ends with 1 errors.\n\n[ERROR] Validation of release number failed. "
+            + "Release number is a patch(10.0.1), but release notes contain 'new' or "
```
rnveach, line 57: Both same here.

releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractPathTestSupport.java
rnveach, line 49: I see usages of `\n` and `System.lineSeparator()`. We should choose one or the other. Beware that if we use JVM's default line separator in production code, Windows uses `\r\n` instead of `\n`.

rnveach (changes requested):

releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractPathTestSupport.java
rnveach: https://ci.appveyor.com/project/Checkstyle/contribution/builds/45632831/job/h7uoii387gp7gu7c#L91
CI agrees.

releasenotes-builder/src/test/java/com/github/checkstyle/MainTest.java
stoyanK7: `isPatch` is checking if the release number ends with a zero `0`. That is why.

releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractPathTestSupport.java
stoyanK7: Wasn't aware of that and was wondering why tests failed there. Thanks! Done.

releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractPathTestSupport.java
stoyanK7: Done.

releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractPathTestSupport.java
stoyanK7: Done.

releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractPathTestSupport.java
stoyanK7: I split them and created `getExecutionFailedMessage(int amountOfErrors)` method.

releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractReleaseNotesTestSupport.java
stoyanK7: Done. I used Nick's suggestion.

releasenotes-builder/src/test/java/com/github/checkstyle/MainTest.java
rnveach: @romani Shouldn't `validateVersion` be checking this?

stoyanK7: validateVersion() calls isPatch()

releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractPathTestSupport.java
rnveach: Not done.

releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractPathTestSupport.java
rnveach: Not done.

releasenotes-builder/src/test/java/com/github/checkstyle/MainTest.java
romani: Let's move this to separate issue  , I will fix it later on

romani: https://github.com/checkstyle/contribution/issues/727

rnveach: Marking resolved

releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractPathTestSupport.java
stoyanK7: Done. Split into `getReleaseIsPatchMessage(String releaseVersion)`.

releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractPathTestSupport.java
stoyanK7: Done. Split into `getReleaseIsMinorMessage(String releaseVersion)`.

rnveach (approved):

PR #722, opened 2022-12-04
Rework test layout

rnveach:
Discussed at https://github.com/checkstyle/contribution/pull/712#issuecomment-1336195814

Reworked and moved code around to build a new testing tier for release notes.

Commit `converts one test in TemplateProcess to the new layout` shows an example of converting an existing test to the new format.

Commit `ensure output location is treated as a directory` makes things easier if output location is always treated as a folder, otherwise we need to manually add `/` when passing it to Main as it doesn't save to the correct location.

### DIFF
--- a/releasenotes-builder/pom.xml
+++ b/releasenotes-builder/pom.xml
@@ -82,6 +82,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>com.github.stefanbirkner</groupId>
+            <artifactId>system-rules</artifactId>
+            <version>1.19.0</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-inline</artifactId>
             <version>${mockito.version}</version>
--- a/releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractPathTestSupport.java
+++ b/releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractPathTestSupport.java
@@ -1,0 +1,124 @@
+///////////////////////////////////////////////////////////////////////////////////////////////
+// checkstyle: Checks Java source code and other text files for adherence to a set of rules.
+// Copyright (C) 2001-2022 the original author or authors.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+///////////////////////////////////////////////////////////////////////////////////////////////
+
+package com.github.checkstyle.internal;
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.util.Locale;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.rules.TemporaryFolder;
+
+import com.github.checkstyle.globals.Constants;
+
+public abstract class AbstractPathTestSupport {
+
+    protected static final String BUG = Constants.BUG_LABEL;
+    protected static final String NEW_FEATURE = Constants.NEW_FEATURE_LABEL;
+    protected static final String NEW_MODULE = Constants.NEW_MODULE_LABEL;
+    protected static final String MISC = Constants.MISCELLANEOUS_LABEL;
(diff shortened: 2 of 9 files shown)


releasenotes-builder/src/test/resources/com/github/checkstyle/githubPageAll.txt
```diff
@@ -2,16 +2,16 @@ Checkstyle 1.0.0 - https://checkstyle.org/releasenotes.html#Release_1.0.0
 
 Breaking backward compatibility:
 
-  #-1 - Title 1
+  #1 - Title 1
```
rnveach, line 5: This is a bug in the prior tests. As we found out when we fail to find issue, it is exception. If it is not an issue commit, it is auto added as miscellaneous and goes under `Other Changes`.
https://github.com/checkstyle/contribution/blob/654a97ee7883f0e46cce0ff11bfb8a6958e3c970/releasenotes-builder/src/main/java/com/github/checkstyle/github/NotesBuilder.java#L175

releasenotes-builder/src/test/resources/com/github/checkstyle/xdocAll.txt
```diff
@@ -17,21 +18,22 @@
...
             Author: Author 3
+            <a href="https://github.com/checkstyle/checkstyle/issues/3">#3</a>
           </li>
         </ul>
       <p>Notes:</p>
         <ul>
           <li>
             Title 4.
             Author: Author 4
-            <a href="https://github.com/checkstyle/checkstyle/issues/4">#4</a>
```
rnveach, line 34: Same bug as previously talked about and everything else in this file.

releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractPathTestSupport.java
```diff
@@ -0,0 +1,124 @@
...
+     */
+    protected void assertFile(String expectedName, String actualName) throws IOException {
+        final String expected =
+                getFileContents(new File(getPath(expectedName)));
+        final String actual = getFileContents(new File(getTempFolder(), actualName))
+                .replace(LocalDate.now()
+                                .format(DateTimeFormatter.ofPattern("dd.MM.yyyy", Locale.US)),
+                        "XX.XX.XXXX");
+
+        Assert.assertEquals("expected " + actualName, expected, actual);
```
rnveach, line 92: The code in this method was moved, but it was also changed because it was specific to xdoc and used for more file. Also changed the assertion to output the file being checked.

releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractReleaseNotesTestSupport.java
```diff
@@ -0,0 +1,201 @@
...
+
+    @Rule
+    public final SystemErrRule systemErr = new SystemErrRule().enableLog().mute();
+    @Rule
+    public final SystemOutRule systemOut = new SystemOutRule().enableLog().mute();
+
+    @BeforeClass
+    public static void setUp() throws IOException {
+        if (!mocked) {
+            mocked = true;
```
rnveach, line 86: As found out in other PR, we can only mock the classes once. This prevents double mocking.

rnveach:
@stoyanK7 Please also share your thoughts.

romani (approved):
It is leap forward.

I wish to avoid fields 
TEST_COMMITS
TEST_ISSUES
At all. But it can be investigated later on .

romani:
We can merge after Stoyan approval

releasenotes-builder/src/test/resources/com/github/checkstyle/githubPageAll.txt
stoyanK7: Noticed that as well when writing tests.

releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractPathTestSupport.java
stoyanK7: Agreed. Would make life easier to see the actual file.

releasenotes-builder/pom.xml
```diff
@@ -81,6 +81,12 @@
             <version>${junit.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.github.stefanbirkner</groupId>
+            <artifactId>system-rules</artifactId>
```
stoyanK7, line 86: Out of curiosity. What is your stance on adding dependencies to checkstyle projects?

rnveach: We don't want to be crazy with dependencies, but if there is an actual need and no other way then we can discuss and include new ones. This dependency is not in main repo as it moved to junit 5 but we are using junit 4 here. Before main repo's upgrade, we use to have this dependency and similar code. I just copied things from the old version of the main repo.

releasenotes-builder/src/test/java/com/github/checkstyle/internal/RevCommitUtil.java
```diff
@@ -30,11 +30,7 @@ public final class RevCommitUtil {
     private RevCommitUtil() {
     }
 
-    public static RevCommit create(String commitMessage) {
-        return create("CheckstyleUser", commitMessage);
-    }
-
-    public static RevCommit create(String author, String commitMessage) {
+    public static RevCommit create(String commitMessage, String author) {
```
stoyanK7, line 33: Thanks for the swap. I found it confusing as well.

stoyanK7:
Looks good to merge!

Issue #733, opened 2022-12-25
releasenotes-builder: no help or usage print by default


rnveach:
````
$ java -jar ~/release/contribution/releasenotes-builder/target/releasenotes-builder-1.0-all.jar

Generation ends with 3 errors.

Remote repository path has not been specified!
Start reference has not been specified!
Release number has not been specified!

$ java -jar ~/release/contribution/releasenotes-builder/target/releasenotes-builder-1.0-all.jar /?

Generation ends with 3 errors.

Remote repository path has not been specified!
Start reference has not been specified!
Release number has not been specified!

$ java -jar ~/release/contribution/releasenotes-builder/target/releasenotes-builder-1.0-all.jar --help
org.apache.commons.cli.UnrecognizedOptionException: Unrecognized option: --help
        at org.apache.commons.cli.DefaultParser.handleUnknownToken(DefaultParser.java:561)
        at org.apache.commons.cli.DefaultParser.handleLongOptionWithoutEqual(DefaultParser.java:383)
        at org.apache.commons.cli.DefaultParser.handleLongOption(DefaultParser.java:334)
        at org.apache.commons.cli.DefaultParser.handleToken(DefaultParser.java:540)
        at org.apache.commons.cli.DefaultParser.parse(DefaultParser.java:712)
        at org.apache.commons.cli.DefaultParser.parse(DefaultParser.java:679)
        at org.apache.commons.cli.DefaultParser.parse(DefaultParser.java:660)
        at com.github.checkstyle.CliProcessor.process(CliProcessor.java:117)
        at com.github.checkstyle.Main.main(Main.java:60)
usage:
...
````

We should print help by default.

romani:
Yes

romani:
fix is merged

PR #735, opened 2022-12-29
Issue 733

rnveach:
Issue #733 

2nd commit moves release specific items out of path test and into the release notes specific test support where there belong. Other minor improvement is adding static to some methods which can be. Eclipse didn't flag them because of the protected modifier.

### DIFF
--- a/releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractReleaseNotesTestSupport.java
+++ b/releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractReleaseNotesTestSupport.java
@@ -52,6 +52,7 @@ import org.mockito.stubbing.Answer;
 import com.github.checkstyle.Main;
 import com.github.checkstyle.git.CsGit;
 import com.github.checkstyle.github.CsGitHub;
+import com.github.checkstyle.globals.Constants;
 import com.github.checkstyle.publishers.TwitterPublisher;
 
 public abstract class AbstractReleaseNotesTestSupport extends AbstractPathTestSupport {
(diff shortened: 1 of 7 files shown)


rnveach:
https://ci.appveyor.com/project/Checkstyle/contribution/builds/45797335/job/ox0n6dmdt47p343n#L827
> Exception in thread "main" javax.xml.stream.XMLStreamException: ParseError at [row,col]:[1,2]
Message: The markup declarations contained or pointed to by the document type declaration must be well-formed.

appveyor looks valid but is not from my changes.

Travis is being travis.

rnveach:
appveyor's XML:
````
````

File: releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractReleaseNotesTestSupport.java [hunk 1]
```diff
@@ -184,7 +194,7 @@ protected Set<RevCommit> getCommits() {
      *
      * @return Copy of all current issues.
      */
-    protected Set<GHIssue> getIssues() {
+    protected static Set<GHIssue> getIssues() {
```
romani, line 197: protected and static is very weird fix.

what force you to do this ? can we aviod this ?
it will make more sense to make it "public static" or move it to some Util.

romani (changes requested):
item:

releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractReleaseNotesTestSupport.java
rnveach: This method was to allow us access to all issues incase a test needed to iterate on them.
`protected` is because this is only called from the classes that inherit this abstract class.
`static` is only being added as improvement as mentioned in first post. We don't need an instance to access this, so usually we make these things static.

I think a util will be weird as it is returning the global field from this class. It is nothing more than a simple getter.

rnveach:
XML format is controlled by us directly at https://github.com/checkstyle/checkstyle/blob/399f1b48b79731d65e824eb927971ef6174c3463/src/main/java/com/puppycrawl/tools/checkstyle/XMLLogger.java#L99 . It has been touched in 8 months.

It stands to reason something changed in patch-diff-report-util for it not to be able to process the XML anymore.

rnveach:
Actually, it is something external to everything. I turned off internet and it was able to process the file. Something external to our repos must have changed and it is failing causing the parsing to fail.

romani:
please rebase to resolve conflict

releasenotes-builder/src/test/java/com/github/checkstyle/internal/AbstractReleaseNotesTestSupport.java
romani: can we just remove "static" ?
and let it be as before -  method of Instance to not share outside details that something is static.

protected imply inheritance, static tell that all inheritence and belonging to class/object is not much relevant.

rnveach: Done.

romani (approved):
ok to merge

PR #745, opened 2022-12-31
Issue #665: adds validation of a configuration being required

rnveach:
Issue #665

````
> groovy .\diff.groovy -r M:\checkstyle -p localtest -l M:\projects-to-test-on.properties

Error: you should specify either 'config', or 'baseConfig', or 'patchConfig'!
Caught: java.lang.IllegalArgumentException: Error: invalid command line arguments!
java.lang.IllegalArgumentException: Error: invalid command line arguments!
````

Program goes to completion with `-c my_check.xml`.

Issue #604, opened 2022-05-01
expand diff.groovy to support 3rd parties


rnveach:
Similar to https://github.com/checkstyle/contribution/issues/340 , I would like diff.groovy to support non-specific 3rd party JARs.

I am splitting this off from sevntu just in-case sevntu's implementation has to be drastically different, or at the very least not conflict since it is slightly higher priority.

This could be done similar to our property file for repos. Give a repo location, build command, and before/after commits or branch names.

Issue #591, opened 2022-01-03
Diff: skip compiling tests


rnveach:
As issue states, we don't need to compile checkstyle's tests when we do an install. We can also skip it when packaging the jar for creating the `patch-diff-report-tool`.

Command is:
`-Dmaven.test.skip=true`

PR #747, opened 2022-12-31
 Issue #273: converts diff.groovy to use Checkstyle's CLI 

rnveach:
Issue #273

Supplemental Checkstyle PR: https://github.com/checkstyle/checkstyle/pull/12576


.ci/travis.sh
```diff
@@ -90,7 +90,7 @@ checkstyle-tester-diff-groovy-regression-single)
   export MAVEN_OPTS="-Xmx2048m"
   groovy ./diff.groovy --listOfProjects projects-to-test-on.properties \
     -pc ../../../checkstyle-tester/diff-groovy-regression-config.xml \
-    -r ../../checkstyle -xm "-Dcheckstyle.failsOnError=false" \
+    -r ../../checkstyle \
```
rnveach, line 93: CLI doesn't have anything similar. It seems we need to build something into groovy.

> If this is true, and Checkstyle reported any violations or errors, the build fails immediately after running Checkstyle

The only way we have to determine if any violations/errors are reported from the CLI is the exit code. Anything else requires adding custom modules to listen for violations.

checkstyle-tester/diff.groovy
```diff
@@ -74,8 +74,8 @@ def getCliOptions(args) {
             + ' as a shorter version to prevent long paths. (optional, default is false)')
         m(longOpt: 'mode', args: 1, required: false, argName: 'mode', 'The mode of the tool:' \
             + ' \'diff\' or \'single\'. (optional, default is \'diff\')')
-        xm(longOpt: 'extraMvnRegressionOptions', args: 1, required: false, 'Extra arguments to pass to Maven' \
-            + 'for Checkstyle Regression run (optional, ex: -Dmaven.prop=true)')
+        xo(longOpt: 'extraRegressionOptions', args: 1, required: false, 'Extra arguments to pass ' \
```
rnveach, line 77: I took `xm` as `extraMvn` so I changed this to `xo` to mean `extraOptions`.

checkstyle-tester/diff.groovy
```diff
@@ -213,10 +213,10 @@ def launchCheckstyleReport(cfg) {
 
     // If "no exception" testing, these may not be defined in repos other than checkstyle
     if (isRegressionTesting) {
-        println "Installing Checkstyle artifact ($cfg.branch) into local Maven repository ..."
+        println "Building Checkstyle artifact ($cfg.branch) ..."
         executeCmd("git checkout $cfg.branch", cfg.localGitRepo)
         executeCmd("git log -1 --pretty=MSG:%s%nSHA-1:%H", cfg.localGitRepo)
-        executeCmd("mvn -e --no-transfer-progress --batch-mode -Pno-validations clean install",
+        executeCmd("mvn -e --no-transfer-progress --batch-mode -Pno-validations clean package -Passembly -Dassembly.skipAssembly=true -Dmaven.test.skip=true",
```
rnveach, line 219: Partial done of #591 .

checkstyle-tester/pom.xml
```diff
@@ -1,71 +0,0 @@
...
-	<artifactId>sample</artifactId>
-	<version>0.0.1-SNAPSHOT</version>
-	<packaging>jar</packaging>
-
-	<properties>
-		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-		<checkstyle-plugin.version>3.1.1</checkstyle-plugin.version>
-		<checkstyle.version>0.0-SNAPSHOT</checkstyle.version>
-		<sevntu-checkstyle.version>1.44.1</sevntu-checkstyle.version>
-		<checkstyle.config.location>https://raw.githubusercontent.com/checkstyle/checkstyle/master/src/main/resources/google_checks.xml</checkstyle.config.location>
```
rnveach, line 17: This config is never really used as shown in #745

checkstyle-tester/pom.xml
```diff
@@ -1,71 +0,0 @@
...
-	<groupId>com.github.sevntu.checkstyle</groupId>
-	<artifactId>sample</artifactId>
-	<version>0.0.1-SNAPSHOT</version>
-	<packaging>jar</packaging>
-
-	<properties>
-		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-		<checkstyle-plugin.version>3.1.1</checkstyle-plugin.version>
-		<checkstyle.version>0.0-SNAPSHOT</checkstyle.version>
-		<sevntu-checkstyle.version>1.44.1</sevntu-checkstyle.version>
```
rnveach, line 16: Sevntu support is completely broken until #340 / #604 . We only supported single mode runs anyways, never supported differences since the sevntu version was never changed.

checkstyle-tester/diff.groovy
```diff
@@ -623,36 +609,35 @@ def getProjectsStatistic(diffDir) {
...
+def runCliExecution(allJar, srcDir, excludes, checkstyleConfig, saveDir,
+                      extraRegressionOptions) {
+    def saveLoc = new File(saveDir)
+    if (!saveLoc.exists()) {
+        saveLoc.mkdirs()
+    }
+
+    println "Running Checkstyle CLI on $srcDir ... with excludes {$excludes}"
+    def cliCommand = "java -jar $allJar -c $checkstyleConfig $srcDir -f xml -o " +
+        getOsSpecificPath("$saveDir", "checkstyle-result.xml") + " -e .git -e .hg"
```
rnveach, line 621: One of the benefits of using the CLI is we now pull in all files and not just `.java`. Because of this we have to exclude `.git` and `.hg` from erroneously processing them. `TreeWalker` will still trim to work only on Java files.

checkstyle-tester/diff.groovy
```diff
@@ -818,7 +805,7 @@ class Config {
             checkstyleCfg: baseConfig,
```
rnveach, line 805: @romani What are these `getCheckstyleToolBaseConfig` for? No one calls them in this script.

.ci/travis.sh
rnveach: I added the reverse option, `continueOnError`. It is inverted because we need the default to be `false` so when we use the option, it is enabled. It follows the POM's default value which was `failsOnError = true` so `continueOnError = false` by default.

checkstyle-tester/diff.groovy
romani: might be dead code - please remove.

romani:
@rnveach , please create PR to main repo to use your branch and validate and action is working to generate diff report.

checkstyle-tester/diff.groovy
rnveach: Not dead code.
https://github.com/checkstyle/contribution/pull/758#issuecomment-1369098505

rnveach:
> create PR to main repo to use your branch 

Done.

checkstyle-tester/diff.groovy
```diff
@@ -628,36 +617,40 @@ def getProjectsStatistic(diffDir) {
...
-    if (extraMvnRegressionOptions) {
-        mvnSite = mvnSite + " "
 
-        if (!extraMvnRegressionOptions.startsWith("-")) {
-            mvnSite = mvnSite + "-"
+    println "Running Checkstyle CLI on $srcDir ... with excludes {$excludes}"
+    def cliCommand = "java -jar $allJar -c $checkstyleConfig $srcDir -f xml -o " +
+        getOsSpecificPath("$saveDir", "checkstyle-result.xml") +
+        " -e " + getOsSpecificPath("$srcDir", ".git") +
+        " -e " + getOsSpecificPath("$srcDir", ".hg")
```
romani, line 631: can we skip all hidden folders ?

checkstyle-tester/projects-for-circle.properties
```diff
@@ -8,12 +8,12 @@ guava|git|https://github.com/google/guava|v28.2||
...
 #openjdk8|hg|http://hg.openjdk.java.net/jdk8/jdk8/jdk/|default|
 #All details at checkstyle/checkstyle#3033: ModalDialogActivationTest till JDK-8166015 ; 'jxc/8073519/**' not compilable by design ; ', jhsdb/**' - checkstyle do not support unicode identifiers
-#openjdk9|hg|http://hg.openjdk.java.net/jdk9/jdk9/jdk/|default|**/test/java/awt/Focus/ModalDialogActivationTest/ModalDialogActivationTest.java,**/test/javax/xml/bind/jxc/8073519/**,**/test/sun/tools/jhsdb/**
+#openjdk9|hg|http://hg.openjdk.java.net/jdk9/jdk9/jdk/|default|/test/java/awt/Focus/ModalDialogActivationTest/ModalDialogActivationTest.java,/test/javax/xml/bind/jxc/8073519/,/test/sun/tools/jhsdb/
 #infinispan|git|https://github.com/infinispan/infinispan|7.2.5.Final||
 #protonpack|git|https://github.com/poetix/protonpack|protonpack-1.7||
 #jOOL|git|https://github.com/jOOQ/jOOL|version-0.9.7||
 #lucene-solr|git|https://github.com/apache/lucene-solr|releases/lucene-solr/5.4.0||
-#tapestry-5|git|https://github.com/apache/tapestry-5|5.4.0|**/archetype-resources/**/*|
+#tapestry-5|git|https://github.com/apache/tapestry-5|5.4.0|/archetype-resources/|
```
romani, line 16: can we keep "**" ? it help to make it clear that it is some part in the middle.

romani (changes requested):
items:

checkstyle-tester/projects-for-circle.properties
rnveach: `**` is not valid regular expression.

romani: yes, I know , but it is still syntax of path matching

checkstyle-tester/diff.groovy
rnveach: How? We cannot identify what is file/folder unless we scan first. We can't be too global as we sometimes work in hidden folders (`.ci-temp`). Also I think there is some issue with relative paths (checkstyle-local has `../../../` which I am still trying to look into where it may not work correctly.

checkstyle-tester/diff.groovy
```diff
@@ -218,11 +220,11 @@ def launchCheckstyleReport(cfg) {
...
     // If "no exception" testing, these may not be defined in repos other than checkstyle
     if (isRegressionTesting) {
-        println "Installing Checkstyle artifact ($cfg.branch) into local Maven repository ..."
+        println "Building Checkstyle artifact ($cfg.branch) ..."
         executeCmd("git checkout $cfg.branch", cfg.localGitRepo)
         executeCmd("git log -1 --pretty=MSG:%s%nSHA-1:%H", cfg.localGitRepo)
-        executeCmd("mvn -e --no-transfer-progress --batch-mode -Pno-validations clean install",
-            cfg.localGitRepo)
+        executeCmd("mvn -e --no-transfer-progress --batch-mode -Pno-validations clean package -Passembly " +
+            "-Dassembly.skipAssembly=true -Dmaven.test.skip=true", cfg.localGitRepo)
```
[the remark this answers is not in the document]
rnveach: `-Pno-validations` atleast skips tidy.
`-Dassembly.skipAssembly=true` skips `maven-assembly-plugin` which builds tar/zip bin/src which is not needed for regression.
`-Dmaven.test.skip=true` skips test resources, compiling, and jar which is also not needed for regression (see https://github.com/checkstyle/contribution/pull/747#discussion_r1059673537 )

checkstyle-tester/diff.groovy
nrmancuso: Why don't we just update regression config to have a filter for these directories? I don't like the idea of hardcoding something that is configurable.

rnveach: Do you mean `projects-to-test-on` files? If we do this, we then have to add it for every project.

nrmancuso: No, in the checkstyle config, by means of BeforeExecutionExclusionFileFilter

checkstyle-tester/projects-for-circle.properties
rnveach: Ok, how? Since it isn't a valid regular expression, it will cause an exception. If we forget it, it just proves how pointless it is.

````
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
          "https://checkstyle.org/dtds/configuration_1_3.dtd">

<module name="Checker">
    <property name="charset" value="UTF-8"/>
    <property name="severity" value="warning"/>
    <property name="haltOnException" value="false"/>

    <module name="TreeWalker">
<module name="Regexp">
  <property name="format" value="Example**"/>
</module>
    </module>
</module>
````

````
Caused by: java.util.regex.PatternSyntaxException: Dangling meta character '*' near index 8
Example**
        ^
    at java.base/java.util.regex.Pattern.error(Pattern.java:2028)
    at java.base/java.util.regex.Pattern.sequence(Pattern.java:2203)
    at java.base/java.util.regex.Pattern.expr(Pattern.java:2069)
    at java.base/java.util.regex.Pattern.compile(Pattern.java:1783)
    at java.base/java.util.regex.Pattern.<init>(Pattern.java:1429)
    at java.base/java.util.regex.Pattern.compile(Pattern.java:1095)
    at com.puppycrawl.tools.checkstyle.utils.CommonUtil.createPattern(CommonUtil.java:100)
    ... 15 more
````

File: checkstyle-tester/diff.groovy (hunk 8 above)
Comment (rnveach) [reply checkstyle-tester/diff.groovy:631]: That is an option, but there is no default configuration file for regression technically. ( see https://github.com/checkstyle/contribution/issues/665 )
It would be up to every config creation to add this.

Comment (nrmancuso) [reply checkstyle-tester/diff.groovy:631]: >It would be up to every config creation to add this.

Yes, it should be up to user to decide what files they want to check (or not check). If somebody runs regression testing over these files, and has a bunch of errors, problem is clear. We point to template in docs, this is enough: https://github.com/checkstyle/contribution/tree/master/checkstyle-tester#report-generation

File: checkstyle-tester/projects-for-circle.properties (hunk 9 above)
Comment (romani) [reply checkstyle-tester/projects-for-circle.properties:16]: https://stackoverflow.com/a/22636142/1015848

Comment (rnveach) [reply checkstyle-tester/projects-for-circle.properties:16]: > Ant-style path patterns matching 

We are using CLI, not Ant or spring.

CLI doesn't support this type of matching, only regular expression.
See https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/Main.java#L785

Comment (romani) [reply checkstyle-tester/projects-for-circle.properties:16]: Yes, let's drop ant syntax.

But can you put in exclude value leading `.*` to show clearly that it is regular regexp.

Comment (rnveach) [reply checkstyle-tester/projects-for-circle.properties]: Done.

File: .ci/validation.sh [hunk 11]
```diff
@@ -95,7 +95,7 @@ checkstyle-tester-diff-groovy-regression-single)
 
 codenarc)
   cd checkstyle-tester
-  ./codenarc.sh . diff.groovy > diff.log && cat diff.log && grep '(p1=0; p2=0; p3=0)' diff.log
+  ./codenarc.sh . diff.groovy > diff.log && cat diff.log && grep '(p1=0; p2=1; p3=0)' diff.log
```
romani, line 98: Can we avoid this ?

checkstyle-tester/diff.groovy
```diff
@@ -628,36 +617,40 @@ def getProjectsStatistic(diffDir) {
...
+    if (extraRegressionOptions) {
+        cliCommand = cliCommand + " "
+
+        if (!extraRegressionOptions.startsWith("-")) {
+            cliCommand = cliCommand + "-"
+        }
+        cliCommand = cliCommand + extraRegressionOptions
+    }
+
+    println(cliCommand)
```
romani, line 651: Let's put some words before command

romani (changes requested):
Items

.ci/validation.sh
rnveach: > Number of parameters in method None.runCliExecution exceeds maximum allowed (5).

Not really. Command needs these number of parameters.

checkstyle-tester/diff.groovy
rnveach: We do have some words at https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/diff.groovy#L636 .

checkstyle-tester/diff.groovy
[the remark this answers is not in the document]
romani: This update shows gaps in our profile
https://github.com/checkstyle/checkstyle/blob/master/pom.xml#L2174
It should skip tidy.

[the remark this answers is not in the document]
romani: We should add all other skips that you need to this profile, as the only goal of this profile is to get Uber jar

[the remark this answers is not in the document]
rnveach: Why do we combine no-validation with assembly? We can keep them separate so we don't have to worry about constantly synching, and we can still combine them both for things like this.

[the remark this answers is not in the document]
romani: We should not combine, I never did this.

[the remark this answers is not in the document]
romani: Do you think we need to make them to not overlap, and use them as combination?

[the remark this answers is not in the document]
rnveach: I think it will be easier to remember and maintain if each profile just have 1 main job. Sevntu does this at https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/.ci/validation.sh#L50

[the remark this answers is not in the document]
romani: Ok let's combine no-validation and assembly.

But assembly should skip tests jaring and assembly plugin. So all 2 parameters should be part of assembly profile.

Does it make sense ?

checkstyle-tester/diff.groovy
romani: It usually bad to print something in out put without context what is it .

romani: We do double print now, second is few lines below.
```
def executeCliCmd(cmd, continueOnError) {
    println "Running command: ${cmd}"
``
So let's remove `println(cliCommand)`

File: .ci/validation.sh (hunk 11 above)
Comment (romani) [reply .ci/validation.sh:101]: lets follow limit. 
lets make method to be `def runCliExecution(allJar, mainClassOptions)`
where mainClassOptions if wrapper object with all data required.

File: checkstyle-tester/diff.groovy (hunk 8 above)
Comment (romani) [reply checkstyle-tester/diff.groovy:631]: filter that we have is working on abslute path, so `\..*` will match to `.ci-temp`.

we can cover this by fansy regexp: `((?!\.ci-temp)\.)+`
https://regex101.com/r/HztiL9/1
https://stackoverflow.com/questions/406230/regular-expression-to-match-a-line-that-doesnt-contain-a-word

File: checkstyle-tester/diff.groovy (hunk 12 above)
Comment (romani) [reply checkstyle-tester/diff.groovy:651]: Please remove this line  is very clear double printing of command.

Comment (rnveach) [reply checkstyle-tester/diff.groovy:651]: Done.

File: .ci/validation.sh (hunk 11 above)
Comment (rnveach) [reply .ci/validation.sh:101]: Should be done.

File: checkstyle-tester/diff.groovy (hunk 8 above)
Comment (romani) [reply checkstyle-tester/diff.groovy]: it think it will work even there is no such folder as ".ci-temp".
so it is config just for our CI execution, in other cases it will not effective.

File: checkstyle-tester/diff.groovy (hunk 10 above)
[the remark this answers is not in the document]
Comment (romani) [reply checkstyle-tester/diff.groovy:227]: @rnveach , is there concerns ?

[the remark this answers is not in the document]
Comment (rnveach) [reply checkstyle-tester/diff.groovy:227]: I combined the 2 profiles into 1 command. `-Pno-validations,assembly`.

I made changes to the profiles in main repo. Let me know if I missed something, otherwise I think this is done.

File: checkstyle-tester/diff.groovy (hunk 8 above)
Comment (rnveach) [reply checkstyle-tester/diff.groovy]: I swapped and made it `-x ((?!\\.ci-temp)\\.)+`.

Comment (rnveach) [reply checkstyle-tester/diff.groovy]: @romani CI failed for this. `checkstyle-tester-diff-groovy-patch` takes place in `.ci-temp`.
https://github.com/checkstyle/contribution/blob/6775a61c51ef392caddd6fef681f5c857bc0220e/.ci/validation.sh#L43-L44

I think we need to keep to explicit excludes.

Comment (rnveach) [reply checkstyle-tester/diff.groovy]: @romani `-x ((?!\.ci\-temp)\.)+` matches everything. I think there is something wrong with your regular expression.

Comment (romani) [reply checkstyle-tester/diff.groovy]: ok, not a big deal, we can improve later on.

**Review by romani — APPROVED** [review]
if CI is green in this repo and it it works well in main repo CI - ok to merge.

File: checkstyle-tester/diff.groovy (hunk 8 above)
Comment (rnveach) [reply checkstyle-tester/diff.groovy]: Reverted code.

File: checkstyle-tester/diff.groovy [hunk 13]
```diff
@@ -823,8 +836,9 @@ class Config {
             checkstyleCfg: baseConfig,
             listOfProjects: listOfProjects,
             destDir: tmpMasterReportsDir,
-            extraMvnRegressionOptions: extraMvnRegressionOptions,
+            extraRegressionOptions: extraRegressionOptions,
             allowExcludes:allowExcludes,
+            continueOnError:continueOnError,
```
rnveach, line 841: Fixed a bug in `continueOnError`, where if new option is not listed in these areas, then we cannot reference them in the main execution as we actually reference them back to here.

Without listed them here, doing `cfg.[option]` in main code always returned `null` and it would not fail anywhere since we were treating it as a boolean. It would just never get enabled.

checkstyle-tester/README.md
```diff
@@ -63,15 +63,18 @@ will be analyzed by Checkstyle during report generation.
...
-would add `--extraMvnRegressionOptions "-Dmaven.site.skip=true"` to `diff.groovy` execution.
+**extraRegressionOptions** (xo) - this option can be used to supply extra command line arguments
+to Checkstyle during the regression run.
 
 **allowExcludes** (g) - this option tells `diff.groovy` to allow paths and files defined in the file
  specified for the `--listOfProjects (-l)` argument to be excluded from report generation
  (optional, default is false). This option is
  used for files that are not compilable or that Checkstyle cannot parse.
 
+**continueOnError** (ce) - whether to fail or continue the Checkstyle' run when it reports a
```
nrmancuso, line 74: `Checkstyle'` -> `Checkstyle`

checkstyle-tester/diff.groovy
```diff
@@ -74,8 +74,10 @@ def getCliOptions(args) {
             + ' as a shorter version to prevent long paths. (optional, default is false)')
         m(longOpt: 'mode', args: 1, required: false, argName: 'mode', 'The mode of the tool:' \
             + ' \'diff\' or \'single\'. (optional, default is \'diff\')')
-        xm(longOpt: 'extraMvnRegressionOptions', args: 1, required: false, 'Extra arguments to pass to Maven' \
-            + 'for Checkstyle Regression run (optional, ex: -Dmaven.prop=true)')
+        ce(longOpt: 'continueOnError', required: false, 'Whether to fail or continue the Checkstyle' \
+            + ' run when it reports a non-zero return code. (optional, default is false)')
+        xo(longOpt: 'extraRegressionOptions', args: 1, required: false, 'Extra arguments to pass ' \
+            + 'for Checkstyle Regression run (optional, ex: -Dprop=true)')
```
nrmancuso, line 80: Please do `Checkstyle Regression run` -> `Checkstyle execution`, it is better to keep it simple. We would still use these options if we weren't "regression" testing, too.

checkstyle-tester/diff.groovy
```diff
@@ -628,36 +625,39 @@ def getProjectsStatistic(diffDir) {
...
-        if (!extraMvnRegressionOptions.startsWith("-")) {
-            mvnSite = mvnSite + "-"
+    println "Running Checkstyle CLI on ${mainClassOptions.srcDir} ... with excludes {${mainClassOptions.excludes}}"
+    def cliCommand = "java -jar $allJar -c ${mainClassOptions.checkstyleConfig} " +
+        "${mainClassOptions.srcDir} -f xml -o " +
+        getOsSpecificPath("${mainClassOptions.saveDir}", "checkstyle-result.xml") +
+        " -e " + getOsSpecificPath("${mainClassOptions.srcDir}", ".git") +
+        " -e " + getOsSpecificPath("${mainClassOptions.srcDir}", ".hg")
+
+    if (mainClassOptions.excludes.length() > 0) {
```
nrmancuso, line 641: Why do we need length check?

checkstyle-tester/diff.groovy
```diff
@@ -628,36 +625,39 @@ def getProjectsStatistic(diffDir) {
...
+    if (!saveLoc.exists()) {
+        saveLoc.mkdirs()
     }
-    if (extraMvnRegressionOptions) {
-        mvnSite = mvnSite + " "
 
-        if (!extraMvnRegressionOptions.startsWith("-")) {
-            mvnSite = mvnSite + "-"
+    println "Running Checkstyle CLI on ${mainClassOptions.srcDir} ... with excludes {${mainClassOptions.excludes}}"
+    def cliCommand = "java -jar $allJar -c ${mainClassOptions.checkstyleConfig} " +
```
nrmancuso, line 635: I can't help but feel that StringBuilder would be faster/easier to read/use less memory, especially with a bunch of excludes.

checkstyle-tester/README.md
nrmancuso: Why do we need this option we are using haltOnException = false and severity = warning?

nrmancuso (changes requested):
Items:

checkstyle-tester/diff.groovy
rnveach: I believe this was because groovy will still create a single entry array when doing the `split` adding an `-x` to the command line with no parameter to exclude.

checkstyle-tester/diff.groovy
rnveach: You want me to switch?

checkstyle-tester/README.md
rnveach: > Why do we need this option we are using haltOnException = false and severity = warning?

See https://github.com/checkstyle/contribution/pull/747#discussion_r1059673349

rnveach: We do have one config used in main repo for regression that has an error set.
https://github.com/checkstyle/checkstyle/blob/d3a5f9a5ba4a1f2df28900dd70f2204c92943521/.ci/no-exception-test.sh#L44-L45

checkstyle-tester/diff.groovy
romani: Keep in mind performance is not an issue here as it is one time execution. Better to keep code easy to read. I don't mind any approach

checkstyle-tester/README.md
rnveach: > Checkstyle' -> Checkstyle

Done.

checkstyle-tester/diff.groovy
rnveach: Done.

checkstyle-tester/README.md
```diff
@@ -63,15 +63,18 @@ will be analyzed by Checkstyle during report generation.
 This option is useful for Windows users where they are restricted to maximum directory depth
 (optional, default is false).
 
-**extraMvnRegressionOptions** (xm) - this option can be used to supply extra command line arguments
-to maven during the Checkstyle regression run.  For example, if you want to skip site generation, you
-would add `--extraMvnRegressionOptions "-Dmaven.site.skip=true"` to `diff.groovy` execution.
+**extraRegressionOptions** (xo) - this option can be used to supply extra command line arguments
+to Checkstyle during the regression run.
```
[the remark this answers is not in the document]
rnveach: Done.

checkstyle-tester/diff.groovy
nrmancuso: It's not a big deal, but we certainly run this more than once (twice for each project). I was thinking more for readability since `+` at the end of the line bugs me. I will leave it up to @rnveach

nrmancuso (changes requested):
Did we test failure at https://github.com/checkstyle/checkstyle/pull/12576, i.e. throw an exception in a check? We need to make sure that we can still fail CI in all cases, too.

checkstyle-tester/diff.groovy
romani: we need to disable trailing "+" rule.
I always thought about groovy as Java script abilities, to write in pure Java but as script. All scripts grows and grow up in applications, so being plain java is good for them

nrmancuso: It is not a rule, this is required by the interpreter

rnveach: @romani Please see the message I posted for you at https://github.com/checkstyle/checkstyle/pull/12725#discussion_r1098907617 . This is groovy itself, and not a style rule.

rnveach:
> Did we test failure at https://github.com/checkstyle/checkstyle/pull/12576, i.e. throw an exception in a check?

It depends on what scenario you are looking for.

Using `<property name="haltOnException" value="false"/>` will ignore the exception and keep going. Similar with `--continueOnError` since this only checks the exit code from Checkstyle.

https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/Main.java#L135
https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/Main.java#L141

I did confirm locally that removing `haltOnException` and `continueOnError`, diff.groovy will stop when an exception occurs. I tested an exception from the config:
````
<module name = "Checker">
    <property name="charset" value="UTF-8"/>

    <!-- do not change severity to 'error', as that will hide errors caused by exceptions -->
    <property name="severity" value="warning"/>

    <module name="RegexpSingleline">
        <property name="format" value="test(.*"/>
    </module>
</module>
````
With output:
````
> groovy diff.groovy -c M:\checkstyleWorkspaceEclipse\contribution\checkstyle-tester\my_check.xml -r Z:\checkstyle -b master -p master -l M:\checkstyleWorkspaceEclipse\contribution\checkstyle-tester\projects-to-test-on.properties

....
Running command: java -jar Z:\checkstyle\target\checkstyle-10.7.1-SNAPSHOT-all.jar -c M:\checkstyleWorkspaceEclipse\contribution\checkstyle-tester\my_check.xml repositories\guava -f xml -o reports\guava\checkstyle-result.xml -e repositories\guava\.git -e repositories\guava\.hg
Exception in thread "main" java.util.regex.PatternSyntaxException: Unclosed group near index 7
test(.*
        at java.base/java.util.regex.Pattern.error(Pattern.java:2027)
        at java.base/java.util.regex.Pattern.accept(Pattern.java:1877)
        at java.base/java.util.regex.Pattern.group0(Pattern.java:3060)
        at java.base/java.util.regex.Pattern.sequence(Pattern.java:2123)
        at java.base/java.util.regex.Pattern.expr(Pattern.java:2068)
        at java.base/java.util.regex.Pattern.compile(Pattern.java:1782)
        at java.base/java.util.regex.Pattern.<init>(Pattern.java:1428)
        at java.base/java.util.regex.Pattern.compile(Pattern.java:1094)
        at com.puppycrawl.tools.checkstyle.checks.regexp.DetectorOptions$Builder.createPattern(DetectorOptions.java:264)
        at java.base/java.util.Optional.map(Optional.java:265)
        at com.puppycrawl.tools.checkstyle.checks.regexp.DetectorOptions$Builder.build(DetectorOptions.java:249)
        at com.puppycrawl.tools.checkstyle.checks.regexp.RegexpSinglelineCheck.beginProcessing(RegexpSinglelineCheck.java:227)
        at com.puppycrawl.tools.checkstyle.Checker.process(Checker.java:218)
        at com.puppycrawl.tools.checkstyle.Main.runCheckstyle(Main.java:415)
        at com.puppycrawl.tools.checkstyle.Main.runCli(Main.java:338)
        at com.puppycrawl.tools.checkstyle.Main.execute(Main.java:195)
        at com.puppycrawl.tools.checkstyle.Main.main(Main.java:130)
Caught: groovy.lang.GroovyRuntimeException: Error: !
groovy.lang.GroovyRuntimeException: Error: !
        at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at diff.executeCliCmd(diff.groovy:698)
        at diff$executeCliCmd$17.callCurrent(Unknown Source)
        at diff.runCliExecution(diff.groovy:660)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at diff$_generateCheckstyleReport_closure4.doCall(diff.groovy:309)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at diff.generateCheckstyleReport(diff.groovy:271)
        at diff$generateCheckstyleReport$10.callCurrent(Unknown Source)
        at diff.launchCheckstyleReport(diff.groovy:233)
        at diff$launchCheckstyleReport$6.callCurrent(Unknown Source)
        at diff.run(diff.groovy:31)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
>
````

rnveach:
All our `no-exception` tests seem to use `continueOnError`.
https://github.com/checkstyle/checkstyle/blob/04fea7164f272af3b73621e4753df5f207033f92/.ci/no-exception-test.sh#L127

This will have to be removed, but there is the one config I pointed out to that uses `error`.
https://github.com/checkstyle/checkstyle/blob/d3a5f9a5ba4a1f2df28900dd70f2204c92943521/.ci/no-exception-test.sh#L44-L45

It looks like `continueOnError` is not a solution, and something in main repo will be needed.

checkstyle-tester/diff.groovy
romani: sad, but ok, lets follow interpreter.

romani:
@nrmancuso , @rnveach , is there easy way to finish this PR ?

rnveach:
The only easy way is to complete everything needed for the transfer ( https://github.com/checkstyle/contribution/pull/747#issuecomment-1426856400 ). This is the only PR in this repo I was working on for this issue.

Issue #453, opened 2020-03-01
Difference report - link to headings


rnveach:
I would like to directly link to headings in the difference report like "Checkstyle configuration report" so I can specifically point to users and say their config is wrong.

It is a very minor thing since that section is still near the top, but I am hoping it will help drive them to see it.

romani:
sounds good.

nrmancuso:
Completed via #749

PR #749, opened 2023-01-01
Issue #453: add links to report headings in patch-diff-report

rnveach:
Issue #453

### DIFF
--- a/patch-diff-report-tool/src/main/resources/header.template
+++ b/patch-diff-report-tool/src/main/resources/header.template
@@ -16,7 +16,7 @@
 			</div>
 
 			<th:block th:if="${config} != null">
-				<h2 a="Checkstyle configuration report">Checkstyle configuration report</h2>
+				<h2><a name="Configuration" href="#Configuration">Checkstyle configuration report</a></h2>
 				<div class="section">
 					<table border="0" class="bodyTable">
 							<tr class="a">
@@ -77,7 +77,7 @@
 			</th:block>
 
 			<div class="section">
-				<h2 a="Summary:">Summary:</h2>
+				<h2><a name="Summary" href="#Summary">Summary:</a></h2>
 				<div class="section">
 					<table border="0" class="bodyTable">
 						<tr class="a">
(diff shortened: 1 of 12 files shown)


rnveach:
Before: http://rveach.no-ip.org/checkstyle/regression/320/openjdk7/index.html
After: http://rveach.no-ip.org/checkstyle/regression/321/openjdk7/index.html#Configuration

The header color change is due to the CSS assigned to the links. I can change this if we want to keep the original red.

rnveach:
http://rveach.no-ip.org/checkstyle/regression/321/pmd/index.html
Above is example with color fix and they are still links.

romani (approved):

rnveach:
@romani Unless you want me to change, the current state of this PR is http://rveach.no-ip.org/checkstyle/regression/321/openjdk7/index.html#Configuration

romani:
it woks good enough.
http://rveach.no-ip.org/checkstyle/regression/321/openjdk7/index.html#A1 still work, so it is good.

nrmancuso (approved):
I am conceptually OK with more links, link-ability is always good.

However, from https://github.com/checkstyle/contribution/issues/453:
>I would like to directly link to headings in the difference report like "Checkstyle configuration report" so I can specifically point to users and say their config is wrong.

In this case, it is probably better to point them to the actual config that is included in the regular diff report (linked xml file in index.html). 

Why don't the reports in this PR have the projects and config files links in index.html? http://rveach.no-ip.org/checkstyle/regression/320/index.html

Example of report with links: https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/0633108_2022211127/reports/diff/index.html

rnveach:
> it is probably better to point them to the actual config 

We did not have those configs at the time of the issue.

> Why don't the reports in this PR have the projects and config files links in index.html? 

My setup uses my scripts. https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/launch_diff.sh
This still runs patch-diff-report, so it is still the same result. It is just my scripts do other things that groovy doesn't like use Checkstyle CLI and allow for resume, pitest, and antlr.

PR #770, opened 2023-02-04
Pull #770: Boost implementation of isIssueOrPull method

stoyanK7:
Part of https://github.com/checkstyle/checkstyle/issues/12717
> Additionally, we should fallback release notes categories to "misc" if something is wrong with parsing a format, we need to update release notes generator.

---
It seems we are already handling this case. The problem was that we were not checking for the hash `#` char in the commit message. We expect to see it as shown here
https://github.com/checkstyle/checkstyle/blob/a144e7fb576f6a875bf5a630190754ca025b7381/src/test/java/com/puppycrawl/tools/checkstyle/internal/CommitValidationTest.java#L74-L75

### DIFF
--- a/releasenotes-builder/src/main/java/com/github/checkstyle/github/CommitMessage.java
+++ b/releasenotes-builder/src/main/java/com/github/checkstyle/github/CommitMessage.java
@@ -61,13 +61,13 @@ import java.util.regex.Pattern;
 
     /**
      * Checks whether commits message is associated with a pull request or an issue.
-     * Commit message which is associated with a pull request or an issue starts with 'Pull'
-     * or 'Issue' prefix.
+     * Commit message which is associated with a pull request or an issue follows one
+     * of the following formats: 'Pull #[number]: [title]' or 'Issue #[number]: [title]'.
      *
      * @return true if commits message is associated with a pull request or an issue.
      */
     public boolean isIssueOrPull() {
-        return message.startsWith("Issue") || message.startsWith("Pull");
+        return message.matches("^(Pull|Issue) #\\d+: .*$");
     }
 
     /**
--- a/releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
+++ b/releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
@@ -288,6 +288,7 @@ public class TemplateProcessorTest extends AbstractReleaseNotesTestSupport {
                 + "LongAndWeExpectItToGetWrappedDueToBeingTooLng", label);
             addIssue(9, CLOSED, "Mock issue title 9 escape @ and @@@@@", label);
             addIssue(10, CLOSED, "Mock issue title 10 escape < > & <&<>&<<", label);
+            addIssue(13, CLOSED, "Mock pull title 13", label);
         }
     }
 
@@ -306,5 +307,11 @@ public class TemplateProcessorTest extends AbstractReleaseNotesTestSupport {
             + "LongAndWeExpectItToGetWrappedDueToBeingTooLng", "Author 11");
         addCommit("Issue #9: Mock issue title 9 escape @ and @@@@@", "Author 12");
         addCommit("Issue #10: Mock issue title 10 escape < > & <&<>&<<", "Author 13");
+        addCommit("Issue 11: Mock issue title 11 missing #", "Author 14");
+        addCommit("Pull 12: Mock pull title 12 missing #", "Author 15");
+        addCommit("miscasmdifs324y23123sda!@.asd12", "Author 16");
+        addCommit("Issue #asd: asd", "Author 17");
+        addCommit("Pull #13: Mock pull title 13", "Author 18");
+        addCommit(".Issue #14: Mock issue title 14", "Author 19");
     }
 }
--- a/releasenotes-builder/src/test/java/com/github/checkstyle/MainTest.java
+++ b/releasenotes-builder/src/test/java/com/github/checkstyle/MainTest.java
@@ -393,4 +393,20 @@ public class MainTest extends AbstractReleaseNotesTestSupport {
         Assert.assertEquals("expected error output", "", systemErr.getLog());
         Assert.assertEquals("expected output", MSG_EXECUTION_SUCCEEDED, systemOut.getLog());
     }
+
+    @Test
+    public void testMistypedCommitMessage() {
+        addCommit("Issue 1: Hello World", "CheckstyleUser");
+        addCommit("Pull 2: Hello World", "CheckstyleUser");
+        addCommit("Issue #asd: asd", "CheckstyleUser");
+
+        runMainContentGenerationAndAssertReturnCode(0,
+            "-releaseNumber", "10.0.1",
+            "-generateAll",
+            "-validateVersion"
+        );
+
+        Assert.assertEquals("expected error output", "", systemErr.getLog());
+        Assert.assertEquals("expected output", MSG_EXECUTION_SUCCEEDED, systemOut.getLog());
+    }
 }
--- a/releasenotes-builder/src/test/resources/com/github/checkstyle/githubPageBreakingCompatibility.txt
+++ b/releasenotes-builder/src/test/resources/com/github/checkstyle/githubPageBreakingCompatibility.txt
@@ -12,6 +12,16 @@ Breaking backward compatibility:
   #8 - Mock issue title 8 thisIssueTitleIsExactly100CharactersLongAndWeExpectItToGetWrappedDueToBeingTooLng
   #9 - Mock issue title 9 escape `@` and `@``@``@``@``@`
   #10 - Mock issue title 10 escape `<` `>` `&` `<``&``<``>``&``<``<`
+  #13 - Mock pull title 13
 
 
 
+<details>
+<summary>Other Changes:</summary>
+<br/>
+  Issue 11: Mock issue title 11 missing # <br/>
+  Pull 12: Mock pull title 12 missing # <br/>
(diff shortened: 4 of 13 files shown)


releasenotes-builder/src/main/java/com/github/checkstyle/github/CommitMessage.java
```diff
@@ -61,13 +61,13 @@ public String getMessage() {
...
-     * Commit message which is associated with a pull request or an issue starts with 'Pull'
-     * or 'Issue' prefix.
+     * Commit message which is associated with a pull request or an issue starts with 'Pull #'
+     * or 'Issue #' prefix.
      *
      * @return true if commits message is associated with a pull request or an issue.
      */
     public boolean isIssueOrPull() {
-        return message.startsWith("Issue") || message.startsWith("Pull");
+        return message.startsWith("Issue #") || message.startsWith("Pull #");
```
rnveach, line 70: @romani Confirm this is what you want.

rnveach (changes requested):

releasenotes-builder/src/main/java/com/github/checkstyle/github/CommitMessage.java
romani: This is not what I meant.
Typo can be any. I would rather wrap whole code in try catch and fallback to misc on exception. And log warn in logs that something weird happened to let maintainers to look, and makes decisions on what to do with it.

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
```diff
@@ -306,5 +306,6 @@ private void createAllCommits() {
             + "LongAndWeExpectItToGetWrappedDueToBeingTooLng", "Author 11");
         addCommit("Issue #9: Mock issue title 9 escape @ and @@@@@", "Author 12");
         addCommit("Issue #10: Mock issue title 10 escape < > & <&<>&<<", "Author 13");
+        addCommit("Issue 11: Mock issue title 11 missing #", "Author 14");
```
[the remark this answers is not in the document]
stoyanK7: Done.

releasenotes-builder/src/main/java/com/github/checkstyle/github/CommitMessage.java
stoyanK7: Should be done. Did some method extraction since the indentation would become too much to bear.

releasenotes-builder/src/main/java/com/github/checkstyle/github/NotesBuilder.java
```diff
@@ -56,6 +56,10 @@ public final class NotesBuilder {
             " https://github.com/%s/issues/%d";
     /** String format pattern for warning if issue is not closed. */
     private static final String MESSAGE_ISSUE_NOT_FOUND = "[WARN] Issue #%d could not be found!";
+
+    /** String format pattern for warning if commit threw an exception during handling.  */
+    private static final String MESSAGE_PROBLEM_HANDLING_COMMIT = "[WARN] An exception occured "
+                            + "while handling commit \"%s\". Assigning miscellaneous label to it.";
```
rnveach, line 62: Where is test showing new message?

rnveach (changes requested):

releasenotes-builder/src/main/java/com/github/checkstyle/github/NotesBuilder.java
stoyanK7: There is no error thrown in the new tests.

rnveach: Yes, it is probably because of the change at https://github.com/checkstyle/contribution/pull/770#discussion_r1096573891 which @romani said was not wanted. I think that code has to be reverted, then your new message will appear.

romani: @stoyanK7, let me know if something needs to be clarified more.
This update blocks our ability to test release.

romani: Please make commit message like `Issue #asd: asd`

stoyanK7: On it.

stoyanK7: Aight, reverting back to `return message.startsWith("Issue") || message.startsWith("Pull");` creates an issue with handling all commits.

Specifically,
https://github.com/checkstyle/contribution/blob/9850b218700203e9c107add7b38a38c5cf7aff8d/releasenotes-builder/src/main/java/com/github/checkstyle/github/NotesBuilder.java#L240-L251

when `commitMessage.getIssueNumber()` is called with let's say `Issue 11: Mock issue title 11 missing #`, it fails the following way since it assumes there is a number between the `#` and the `:`:

Screenshot from 2023-02-08 09-49-33

stoyanK7: Same goes with  `Issue #asd: asd`. Wouldn't it be a good idea to change the implementation of `isIssueOrPull()` to use regex to prevent this and check for `^(Pull|Issue) #\d+:.*$`?

Screenshot from 2023-02-08 10-19-46

stoyanK7: All calls to `getIssueNumber()` are guarded by checking if `isIssueOrPull()`.

romani: Yes, please boost isIssueOrPull() by all means.
Catching exception still good, to not deal with any unfortunate edge case

stoyanK7: @romani By above method, you mean `getCommitsForIssue()`?

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
```diff
@@ -288,6 +288,7 @@ private void createAllIssues(String... labels) {
                 + "LongAndWeExpectItToGetWrappedDueToBeingTooLng", label);
             addIssue(9, CLOSED, "Mock issue title 9 escape @ and @@@@@", label);
             addIssue(10, CLOSED, "Mock issue title 10 escape < > & <&<>&<<", label);
+            addIssue(13, CLOSED, "Mock pull title 13", label);
```
stoyanK7, line 291: @rnveach Since I added a commit starting with `Pull #13:`, I had to add a corresponding issue to get rid of the warning that it does not exist. However, this does not feel right, since I am creating a pull request and not an issue technically. For now, it works as intended, but I think we can extend `AbstractReleaseNotesTestSupport` with `addPullRequest()` for instance.

releasenotes-builder/src/main/java/com/github/checkstyle/github/NotesBuilder.java
stoyanK7: Done. Boosted implementation and added one more test case.

romani (approved):
Ok to merge

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
rnveach: The reason it works is because to the GH api it is all the same. https://github.com/checkstyle/contribution/blob/2617fbafa32323ef3c3a4b25172388451177c8c1/releasenotes-builder/src/main/java/com/github/checkstyle/github/NotesBuilder.java#L148

Anything like `addPullRequest` will just be a wrap to `addIssue`.

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
```diff
@@ -306,5 +307,10 @@ private void createAllCommits() {
             + "LongAndWeExpectItToGetWrappedDueToBeingTooLng", "Author 11");
         addCommit("Issue #9: Mock issue title 9 escape @ and @@@@@", "Author 12");
         addCommit("Issue #10: Mock issue title 10 escape < > & <&<>&<<", "Author 13");
+        addCommit("Issue 11: Mock issue title 11 missing #", "Author 14");
+        addCommit("Pull 12: Mock pull title 12 missing #", "Author 15");
+        addCommit("miscasmdifs324y23123sda!@.asd12", "Author 16");
+        addCommit("Issue #asd: asd", "Author 17");
+        addCommit("Pull #13: Mock pull title 13", "Author 18");
```
rnveach, line 314: Please add a test case that shows the `^` is needed for the regular expression you changed.

rnveach (changes requested):

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
stoyanK7: Alright, thanks for the clarification!

releasenotes-builder/src/test/java/com/github/checkstyle/templates/TemplateProcessorTest.java
stoyanK7: Done.
`.Issue #14: Mock issue title 14`

rnveach (approved):

PR #821, opened 2023-12-10
minor: add openjdk21 to projects file for diff report generation

nrmancuso:


rnveach:
@nrmancuso ping. What blocks this?

nrmancuso:
> @nrmancuso ping. What blocks this?

I think this will make diff report execution take too long/too much memory. I need to reevaluate this. We should probably just replace 17 with 21, if it’s one or the other.

romani:
Only one jdk code is required for execution. Keep latest with exclusion on files that we can not parse yet.

nrmancuso:
> Only one jdk code is required for execution. Keep latest with exclusion on files that we can not parse yet.

Agreed, they do a good job of keeping old inputs when possible, so we really shouldn’t miss anything by always using the latest version.

mahfouz72:
@rnveach @nrmancuso @romani 

I think this got missed please replace jdk17 with 21. As you said old inputs aren't removed so we won't miss it we just gain some new inputs related to the new syntax.

why do we have this huge exclude list?

nrmancuso:
> @rnveach @nrmancuso @romani
> 
> I think this got missed please replace jdk17 with 21. As you said old inputs aren't removed so we won't miss it we just gain some new inputs related to the new syntax.
> 
> why do we have this huge exclude list?

Maybe we don't need this anymore with our new TreeWalker property :)

romani:
Only one jdk code is required for execution, jdk17 can be removed.

rnveach:
Exclude list is needed for no violation/exception runs.

nrmancuso:
Closed in favor of https://github.com/checkstyle/contribution/pull/878

PR #828, opened 2024-01-29
Update README.md

Rayyan925:


Rayyan925:
I have updated this file

nrmancuso:
@Rayyan925 we will not be accepting this change. We mostly stick to typical GitHub README format/ markdown convention with headers like this. Also - we will not consider contributions with no context/reason. If you would like to contribute to Checkstyle, please take a look at https://github.com/checkstyle/checkstyle/wiki/To-New-Contributor.

Issue #757, opened 2023-01-02
Update codenarc to latest version


nrmancuso:
Noticed at https://github.com/checkstyle/contribution/pull/746/files#r1060109071, we are using an old version of codenarc. Latest release is 3.1.0, we are using 1.4/1.5 right now.

https://github.com/CodeNarc/CodeNarc/releases

Edit: we cannot use the latest due to older groovy version used in CI and the groovy syntax used in our scripts. We have settled on 2.2.0

nrmancuso:
Closed via https://github.com/checkstyle/contribution/pull/836

nrmancuso:
@MANISH-K-07 do we need to do the same in the main repo?

MANISH-K-07:
> @MANISH-K-07 do we need to do the same in the main repo?

That was the idea, yes. I was waiting for this PR to get merged. Now, I will bump release in main repo

We are still using the older version `v1.4`
From checkstyle/[codenarc.groovy](https://github.com/checkstyle/checkstyle/blob/master/.ci/codenarc.groovy) : 

```
@Grapes(
        @Grab(group='org.codenarc', module='CodeNarc', version='1.4')
)
```

We might expect some violations as part of release bump....

MANISH-K-07:
@nrmancuso ,

I have opened an issue in the main repo : https://github.com/checkstyle/checkstyle/issues/14556 (Issue)
Linked a PR that aims to close the above issue : https://github.com/checkstyle/checkstyle/pull/14558 (PR)

Please review....

PR #835, opened 2024-02-22
Issue #757: Bumped codenarc  to latest version v3.4.0

MANISH-K-07:
Closes Issue #757

### DIFF
--- a/checkstyle-tester/codenarc.sh
+++ b/checkstyle-tester/codenarc.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env groovy
 @Grapes(
-@Grab('org.codenarc:CodeNarc:1.5')
+@Grab('org.codenarc:CodeNarc:3.4.0')
 )
 @GrabExclude('org.codehaus.groovy:groovy-xml')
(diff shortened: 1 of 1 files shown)


MANISH-K-07:
The error is : 

```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.codehaus.groovy.reflection.CachedClass (file:/usr/share/groovy/lib/groovy-2.4.21.jar) to method java.lang.Object.finalize()
WARNING: Please consider reporting this to the maintainers of org.codehaus.groovy.reflection.CachedClass
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
java.lang.AbstractMethodError: Receiver class org.codenarc.CodeNarc does not define or inherit an implementation of the resolved method 'abstract void setProperty(java.lang.String, java.lang.Object)' of interface groovy.lang.GroovyObject.
```

@nrmancuso ,
I am not sure if this is true, but I don't think a workaround to the above error exists with the current groovy version that checkstyle uses ( `2.4.x stream` )

PR #867, opened 2024-05-27
Issue #866: Convert project-to-test into yaml

### LINKED ISSUE #866: checkstyletester: Yaml firmat for project list
We have now very weird format of project list.
https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/projects-to-test-on.properties
It was quick and old decision.
Let's do yaml format for it.

relentless-pursuit:
Resolves #866


rnveach:
There is a file conflict.

Also, if we are making this a YAML file, I would expect the extension to be `.yml`.

I also think this type of change will be pretty wide. Our regression groovy will need to be modified to now work with this new file format.

I am not sure if we need to update our GH action in main repo that allows users to use custom test property file.

If we rename the file the groovy change will have to be done in this same PR. You may want to consider just creating a whole new file, and then do the groovy update and property deletion in another file.

rnveach:
Final note from me, I dislike commenting out system, if that even works for YAMLs.

Can we switch to something like enabled/disabled instead?

This will also impact our project specific regression in main repo. We comment out individual lines that match project name. This will make enabling/disabling specific projects a bit more complex. I think enable/disable option is the best in this case, but we will need a way to locate and enable project specifics.

nrmancuso:
> Can we switch to something like enabled/disabled instead?

Second this. We have no need for commenting out of we have this field

romani:
Please send PR to https://github.com/romani/contribution

PR #785, opened 2023-06-01
minor: New Projects added for Diff regression

Kevin222004:
minor: New Projects added for Diff regression

PR #594, opened 2022-01-20
config: add openjdk17 project to projects-to-test-on-for-github-action

nmancus1:
Adds openjdk17 to projects file for report generation in main repo by github action.

Successful usage of this file at https://github.com/checkstyle/checkstyle/pull/11221#issuecomment-1017163882

This can’t be merged until https://github.com/checkstyle/checkstyle/pull/11100

PR #584, opened 2021-12-04
minor: remove excludes(related to checkstyle#10956)

nmancus1:
Related to comments at https://github.com/checkstyle/checkstyle/pull/10956#issuecomment-970332760

PR #878, opened 2024-07-14
minor: add openjdk21 and remove openjdk17 from projects file for report generation

mahfouz72:


mahfouz72:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-site-plugin:3.12.1:site (default-site) on project sample: Error generating maven-checkstyle-plugin:3.1.1:checkstyle report: Failed during checkstyle execution: There is 1 error reported by Checkstyle 10.18.0-SNAPSHOT with my_check.xml ruleset. -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-site-plugin:3.12.1:site (default-site) on project sample: Error generating maven-checkstyle-plugin:3.1.1:checkstyle report
```

please help with this I don't understand the failure. I need this update to use the default list. We should have openjdk21 in the default projects now.

rnveach:
Provide us a link to what is being executed before this diff.groovy call so we can see the context of the job.

rnveach:
I meant the source code. You need know what is the context of this job and what it is trying to do to understand what the failure is.

mahfouz72:
This job is trying to generate reports right? what I see is that it clones checkstyle repo and the report generation failed with this message https://github.com/checkstyle/contribution/pull/878#issuecomment-2228784742. So what source code do you refer to?

rnveach:
> So what source code do you refer to?

Looking at https://app.circleci.com/pipelines/github/checkstyle/contribution/363/workflows/cbafbb70-b92e-446d-aeec-82ca8f4ba40e/jobs/2516

each section has a title. The title of section with the job failure is `./.ci/validation.sh checkstyle-tester-diff-groovy-patch`. That is the job being executed, so we will need to look at it's source to see what it is doing and trying to understand the failure.

mahfouz72:
I tried to run this script on local, build was successful
```
Creation of configuration report is started.
Creation of diff html site is started.
Creation of the result site succeed.
patch-diff-report-tool execution finished.
Running command: java -jar /home/mahfouz/contribution/patch-diff-report-tool/target/patch-diff-report-tool-0.1-SNAPSHOT-jar-with-dependencies.jar --patchReport reports/patch-branch/checkstyle-sonar/checkstyle-result.xml
                        --output reports/diff/checkstyle-sonar --patchConfig my_check.xml --baseReport reports/master/checkstyle-sonar/checkstyle-result.xml --baseConfig my_check.xml
patch-diff-report-tool execution started.
XML parsing is started.
Creation of configuration report is started.
Creation of diff html site is started.
Creation of the result site succeed.
patch-diff-report-tool execution finished.
Diff report generation finished ...
Starting creating report summary page ...
Creating report summary page finished...
```

mahfouz72:
CI is green idk why it failed before please consider merging

rnveach:
Can you provide the link to the source of the job?

mahfouz72:
https://github.com/checkstyle/contribution/blob/fb4a65b361fd6c6887444eb0e89a9191ec89d789/.ci/validation.sh#L35-L46

rnveach:
So this test regression on checkstyle's master against itself and runs regression on checkstyle project only. The default config file is really set up to produce no violations. It is basically a no harms test of the diff groovy.

It uses `projects-to-test-on.properties` but you modified `projects-to-test-on-for-github-action.properties`, so yes, this test has no connection to your changes.

It didn't specify what the error was, so we won't know what the issue was. So its failure does seem unrelated.

rnveach:
Since this says for github actions, I assume this means in main repo. Is this file and openjdk17 used anywhere in main repo that we have switch it out once this is merged?

mahfouz72:
I don't think so. This file is not in the main repo. Updating this file is enough. It is used in diff-report.yml 
https://github.com/checkstyle/checkstyle/blob/d335962fa743408b6823236cd5c352eb7ea7965a/.github/workflows/diff-report.yml#L18-L20

I just don't know why we have those files:
- https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/github-action-projects1.properties
- https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/github-action-projects2.properties

They are not used anywhere why are we keeping them?

rnveach:
> This file is not in the main repo. Updating this file in this PR is enough. 

It is only enough if the main repo does not make use of this file or this property you changed. If we made use of this in main repo, we would most likely get a failure after merge since openjdk17 no longer exists.

> It is used in diff-report.yml

So this is only used for default configuration for regression. Users can supply an override and all projects in this file should be uncommented.

rnveach:
@romani @nrmancuso 
> don't know why we have those files:

Are we keeping these files for some reason? They were added over a year ago and they just say `minor: New Projects added for Diff regression`.

checkstyle-tester/projects-to-test-on-for-github-action.properties
```diff
@@ -40,5 +40,4 @@ nbia-dcm4che-tools|git|https://github.com/thprakash/nbia-dcm4che-tools|c3591e6f0
 # RequireThis usage
 spring-integration|git|https://github.com/spring-projects/spring-integration|main|
 
-# openjdk 17 requires lots of excludes, list here should be consistent with file filters at https://github.com/checkstyle/checkstyle/blob/master/.ci/openjdk17-excluded.files
```
rnveach, line 43: @nrmancuso This line says it must have excludes consistent with main repo. Are you sure we can remove all excludes?

rnveach (approved):

checkstyle-tester/projects-to-test-on-for-github-action.properties
mahfouz72: Related discussions:

https://github.com/checkstyle/checkstyle/pull/10956

https://github.com/checkstyle/contribution/pull/594

romani:
Files are added in this PR
https://github.com/checkstyle/contribution/pull/785#pullrequestreview-1455790371

Unfortunately no reference to why we need this.
Sounds like there were some testing activities on extended list of projects, and most likely me asked to preserve list of projects.
But I don't see anything reference them in PR descriptions , looks like they never used.

checkstyle-tester/projects-to-test-on-for-github-action.properties
mahfouz72: I tried to catch up with discussions

We used to have an exclude list (non-compilable file) for all projects in this file. [but we then removed all excludes from all projects](https://github.com/checkstyle/contribution/pull/584#issuecomment-986359760) because it turned out that these excluded files were ignored. We removed them to be consistent with previous reports. 

After that, [we added support for exclude option with `--allowExcludes`](https://github.com/checkstyle/checkstyle/pull/10956). But we never returned the exclude list to the projects list. 

Then [we added this openjdk17 with this huge exclude list](https://github.com/checkstyle/contribution/pull/594). 

I can confirm that this file `(projects-to-test-on-for-github-action.properties)` isn't used by any other task like no-exception. It is only used as a `DEFAULT_PROJECTS_LINK` in diff-report.yml. So, removing this exclude list and openjdk17 will not affect any other task.

now the question should we add an exclude list for the new openjdk21? If yes, then why don't we add an exclude list for the other projects?

> Should we add an exclude list for the new openjdk21?

I don't think we gain any thing if we add it. we have 2 cases :

- non-compilable file and can't be parsed by us: This will get ignored due to the `skipFileOnJavaParseException` new treewalker property 

- non-compilable file but can be parsed ([example](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/f5590ec_2022232703/reports/diff/openjdk17/xref/home/runner/work/checkstyle/checkstyle/contribution/checkstyle-tester/repositories/openjdk17/test/langtools/tools/javac/diags/examples/DuplicateTotalPattern.java.html#L31)): This will have no parse error so will not be skipped and it will appear in the report. Is this acceptable?  for me, I don't think we will lose anything by not excluding those files On the contrary they might help ensure regression ([example for a  non-compilable file that shows that the check update is as expected](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/f5590ec_2022232703/reports/diff/openjdk17/index.html#A1)  from [this PR](https://github.com/checkstyle/checkstyle/pull/11221#issue-1108696113)) 
 > I am not sure if we should add files filters for these files, though, since they are how I confirmed logic of check update.

mahfouz72: @nrmancuso 
Please correct me if anything is wrong I tried to go quickly over the blame of this file. You must have strong opinion because all these PR are created by you :)

mahfouz72:
> Unfortunately no reference to why we need this. 

Then we should have a good reference in this PR about why we did this update before merge to help future us :)

checkstyle-tester/projects-to-test-on-for-github-action.properties
romani: >non-compilable file but can be parsed ([example](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/f5590ec_2022232703/reports/diff/openjdk17/xref/home/runner/work/checkstyle/checkstyle/contribution/checkstyle-tester/repositories/openjdk17/test/langtools/tools/javac/diags/examples/DuplicateTotalPattern.java.html#L31)): This will have no parse exception so will not be skipped and it will appear in the report. Is this acceptable? for me, I don't think we will lose anything by not excluding those

Thanks a lot for such detailed post.
There is a nuance. If file is not compile able by javac, but parse able by checkstyle - target Check will execute and produce some violations or no violations.
It will be very confusing, as in report it will looks like diff in behavior and will be red flag for all reviewers. It will be not easy to catch a fact that this file is not legitimate to run on.

Removal of excludes is prone to cause bunch of false investihations for author of PR or bad requests from maintainers to investigate differences.

I think we need to keep exclude list for open jdk.

romani (changes requested):
I placed reason to keep excludes in comment above

checkstyle-tester/projects-to-test-on-for-github-action.properties
mahfouz72: I forgot to ask a question in my post.

> file is not compile able by javac, but parse able by checkstyle

How do we even know these files?

mahfouz72: Okay looks like this huge exclude list for non-parseable files by CS only. It didn't include the first case (file is not compile able by javac, but parse able by checkstyle) I discovered this in a very recent [report](https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/b2743e3_2024192510/reports/diff/openjdk17/index.html#A563) it include a non compile able but parse able file. I think it makes sense we have no way to discover such files. In this case, the excludes is for non-parseable files only which can be completely ignored by the treewalker property 

@romani please reconsider,  can we find any way to know not compilable by javac, but parse able by checkstyle files

romani: > How do we even know these files?

it was hard work of @nrmancuso .

>  can we find any way to know not compilable by javac, but parse able by checkstyle files

only by pratice and spotting something weird in diff report.
We can make a script to find all, but I am not sure that it is reasonable to do this at summer.

> In this case, the excludes is for non-parseable files only which can be completely ignored by the treewalker property

should be skipped completely for diff report.
Treewalker can handle all, this list is for known files that can give weird behavior and confuse readers of diff report.

romani: @nrmancuso , please step in.

rnveach:
@mahfouz72 Since there is too much going on in this issue to possibly get lost, please move https://github.com/checkstyle/contribution/pull/878#issuecomment-2229543870 to a new issue.

We should have an issue if we are going to remove them anyways, otherwise, we will document why we need the files and possibly add the reason to the file(s).

checkstyle-tester/projects-to-test-on-for-github-action.properties
nrmancuso: These exclusions are to avoid having to analyze files that don't matter (noncompilable) or that we don't support parsing. It is generated by bash/python from [here](https://github.com/checkstyle/checkstyle/blob/f5c492af3fde6bd417053e2f5a0d187767b3551b/config/projects-to-test/openjdk21-excluded.files#L1).

Let's just generate these and move on, we can debate the value of having these in another issue

romani: Look like @nrmancuso , agree to keep this list but content might be different.

mahfouz72:
> We should have an issue if we are going to remove them anyways, otherwise, we will document why we need the files and possibly add the reason to the file(s).

https://github.com/checkstyle/contribution/issues/880

checkstyle-tester/projects-to-test-on-for-github-action.properties
mahfouz72: @nrmancuso @romani done please approve
issue : https://github.com/checkstyle/contribution/issues/881

nrmancuso (approved):

romani (approved):
Ok to merge.

Looks like we regenerated exclude list that is good.

romani:
test-configs will be updated to have openjdk21 https://github.com/checkstyle/test-configs/issues/124

Issue #957, opened 2025-08-04
Upgrade all scripts to groovy 3.0.17 or 5.x


AmitKumarDeoghoria:
Detected at https://github.com/checkstyle/checkstyle/pull/17267/files

We need to upgrade scripts to newest groovy, grovy 4 is also fine

Groovy 5 is released, https://groovy.apache.org/download.html .

romani:
Better to use groovy 4 , to not returning back to this question in several years.

PR #958, opened 2025-08-04
Issue #957: Updated Docker Script to latest Groovy version

AmitKumarDeoghoria:
part of : #957


AmitKumarDeoghoria:
why is it failing?  Time limits?

romani:
all green , as there will be prove that this docker container is working, we can merge.

Issue #963, opened 2025-08-16
Migrate JDK to 21


AmitKumarDeoghoria:
Main Project getting Migrated to JDK21: https://github.com/checkstyle/checkstyle/issues/17429
 we need to migrate all in this repo to jdk21

romani:
major update was done at https://github.com/checkstyle/contribution/pull/1011

test will be updated at https://github.com/checkstyle/contribution/issues/1012

PR #964, opened 2025-08-16
Issue #963: Migrate Docker Script to jdk21

AmitKumarDeoghoria:
part of : #963


romani:
good, we need a prove that it works

